PR #338, opened 2015-01-15
BFV bug

jrperritt:
If a server is created via boot from volume, it won't have a map[string]interface{} type for the `image` field; it will instead be of type `string`. This PR allows for an `image` attribute of type string when retrieving a server created via bfv.

### DIFF
--- a/openstack/compute/v2/servers/results.go
+++ b/openstack/compute/v2/servers/results.go
@@ -1,6 +1,8 @@
 package servers
 
 import (
+	"reflect"
+
 	"github.com/mitchellh/mapstructure"
 	"github.com/rackspace/gophercloud"
 	"github.com/rackspace/gophercloud/pagination"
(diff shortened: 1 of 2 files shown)


openstack/compute/v2/servers/results.go
```diff
@@ -20,8 +22,26 @@ func (r serverResult) Extract() (*Server, error) {
...
-	return &response.Server, err
+	config := &mapstructure.DecoderConfig{
+		DecodeHook: func(from reflect.Kind, to reflect.Kind, data interface{}) (interface{}, error) {
+			if (from == reflect.String) && (to == reflect.Map) {
+				return map[string]interface{}{}, nil
+			}
+			return data, nil
+		},
+		Result: &response,
+	}
```
smashwilson: We could reduce the duplication between `Extract` and `ExtractServer` by extracting `config` to a package-private var:

```go
var serverDecoderConfig = &mapstructure.DecoderConfig{
  // ...
}

decoder, err := mapstructure.NewDecoder(serverDecoderConfig)
// etc.
```

smashwilson:
Mostly out of curiosity: do we have an example of a response body that uses a string from its image? Is it an empty string?

This does make me wonder if there are any other server attributes that can conditionally break the schema like this...

jrperritt:
In the tests that I ran, the `image` attribute has the empty string as its value. I'd guess because BFV doesn't use an image to create the server.

openstack/compute/v2/servers/results.go
jrperritt: I considered that, but the `Result` field has different types for each of the functions.

smashwilson: Oh, right.

Maybe just the `DecodeHook`, then?

jrperritt: Sure

smashwilson:
:+1: :shipit:

PR #339, opened 2015-01-23
OpenStack and Rackspace CDN Service Support

jrperritt:
OpenStack Ref: http://docs.cloudcdn.apiary.io/#services
Rackspace Ref: http://docs.rackspace.com/cdn/api/v1.0/cdn-devguide/content/Overview.html

This PR adds OpenStack and Rackspace support for the new CDN service.


openstack/cdn/v1/services/results.go
```diff
@@ -0,0 +1,197 @@
...
+	return response.Services, err
+}
+
+// CreateResult represents the result of a Create operation.
+type CreateResult struct {
+	gophercloud.Result
+}
+
+// Extract is a method that extracts the location of a newly created service.
+func (r CreateResult) Extract() (string, error) {
```
jrperritt: The only interesting thing that `services.Create` (and `services.Update`) returns is the URL location of the service in the `Location` header. The service ID is embedded in the location URL, which is why the Get/Update/Delete `service` functions let users pass in a `string` that can be a URL or a service ID.

openstack/cdn/v1/base/requests.go
```diff
@@ -0,0 +1,30 @@
...
+func Get(c *gophercloud.ServiceClient) GetResult {
+	var res GetResult
+	_, res.Err = perigee.Request("GET", getURL(c), perigee.Options{
+		MoreHeaders: c.AuthenticatedHeaders(),
+		Results:     &res.Body,
+		OkCodes:     []int{200},
+	})
+	return res
+}
+
```
smashwilson: Do we also want to support the `/health` and `/health/{subsystem}` calls here?

Looking at the responses they seem mostly useful for internal diagnostics, so I could also see the argument for leaving them out. For completeness' sake it might be worth having them, though. "In a later PR" would also be a valid option.

openstack/cdn/v1/base/requests_test.go
```diff
@@ -0,0 +1,45 @@
...
+        },
+        "hints": {
+            "allow": [
+                "GET"
+            ],
+            "formats": {
+                "application/json": {}
+            }
+        }
+    }`,
```
smashwilson: Hmm. The home document isn't parsed the whole way? This looks like it's left mostly unparsed as a `map[string]string`.

openstack/cdn/v1/flavors/requests_test.go
```diff
@@ -0,0 +1,93 @@
...
+
+    expected := []Flavor{
+      Flavor{
+        ID:   "europe",
+        Providers: []Provider{
+          Provider{
+            Provider: "Fastly",
+            Links: []gophercloud.Link{
+              gophercloud.Link{
+                Href: "http: //www.fastly.com",
```
smashwilson: There's an extra space here (and in the corresponding fixture response). :point_up: No big deal, it just looks a little odd.

openstack/cdn/v1/flavors/requests_test.go
```diff
@@ -0,0 +1,93 @@
...
+
+    th.CheckDeepEquals(t, expected, actual)
+
+    return true, nil
+  })
+  th.AssertNoErr(t, err)
+
+  if count != 1 {
+    t.Errorf("Expected 1 page, got %d", count)
+  }
```
smashwilson: This can be:

```go
th.CheckEquals(t, 1, count)
```

openstack/cdn/v1/flavors/results.go
```diff
@@ -0,0 +1,71 @@
...
+import (
+	"github.com/mitchellh/mapstructure"
+	"github.com/rackspace/gophercloud"
+	"github.com/rackspace/gophercloud/pagination"
+)
+
+// Provider represents a provider for a particular flavor.
+type Provider struct {
+	// Specifies the name of the provider. The name must not exceed 64 bytes in
+	// length and is limited to unicode, digits, underscores, and hyphens.
```
smashwilson: :+1: for being explicit about limitations in the struct docs.

openstack/cdn/v1/services/requests.go
```diff
@@ -0,0 +1,319 @@
...
+	Value map[string]interface{} `json:"value,omitempty"`
+}
+
+// ToCDNServiceUpdateMap casts an UpdateOpts struct to a map.
+func (opts UpdateOpts) ToCDNServiceUpdateMap() ([]map[string]interface{}, error) {
+	s := make([]map[string]interface{}, len(opts))
+
+	for i, opt := range opts {
+		if opt.Op == "" {
+			return nil, no("Op")
```
smashwilson: You should also validate that `opt.Op` is one of `Add`, `Remove` or `Replace` here, because technically a client can cast any string to an `Op`.

The *other* option is to change `type Op string` to `type Op struct { name string }`, which prevents casting, but would require slightly more code to pull the `name` field out here - your choice.

openstack/cdn/v1/services/requests_test.go
```diff
@@ -0,0 +1,333 @@
...
+
+	err := List(fake.ServiceClient(), &ListOpts{}).EachPage(func(page pagination.Page) (bool, error) {
+		count++
+		actual, err := ExtractServices(page)
+		if err != nil {
+			t.Errorf("Failed to extract services: %v", err)
+			return false, err
+		}
+
+		expected := []Service{
```
smashwilson: Haha. Aren't giant structs fun?

openstack/cdn/v1/services/results.go
smashwilson: :+1: Cool. This is a great reason to have the `.Extract()` layer to abstract this kind of thing from users, too :grinning:

rackspace/cdn/v1/base/delegate_test.go
```diff
@@ -0,0 +1,46 @@
...
+        },
+        "hints": {
+            "allow": [
+                "GET"
+            ],
+            "formats": {
+                "application/json": {}
+            }
+        }
+    }`,
```
smashwilson: Same as the OpenStack comment about this test: it's weird to me how this document is being parsed only a single level down.

rackspace/cdn/v1/flavors/delegate_test.go
```diff
@@ -0,0 +1,93 @@
...
+
+		th.CheckDeepEquals(t, expected, actual)
+
+		return true, nil
+	})
+	th.AssertNoErr(t, err)
+
+	if count != 1 {
+		t.Errorf("Expected 1 page, got %d", count)
+	}
```
smashwilson:
```go
th.CheckEquals(t, 1, count)
```

smashwilson:
@jrperritt Looking great :grinning: I left a few comments - mostly questions, nothing showstopping except *maybe* the way that `HomeDocument` is being returned.

openstack/cdn/v1/base/requests.go
jrperritt: Ah, you read my mind. This PR just handles the operations that OpenStack and Rackspace share.

smashwilson: :+1: Works for me.

rackspace/cdn/v1/base/delegate_test.go
jrperritt: You're right. Good catch. I'll fix that.

openstack/cdn/v1/services/requests_test.go
jrperritt: A necessary evil of statically-typed languages, I suppose :wink:

smashwilson:
:metal: :shipit:

PR #340, opened 2015-01-23
Omit the Content-Type header on an objects.Create call unless intentionally provided

### LINKED ISSUE #336: By default, Object Storage attempts to interpret its io.Reader as a JSON object
If you don't provide a `ContentType` manually, Perigee will [default it to `application/json`](https://github.com/racker/perigee/blob/master/api.go#L45-L47) and then [attempt to marshal the `io.Reader` as a JSON object and upload that instead](https://github.com/racker/perigee/blob/master/api.go#L49-L57). If you're just providing an open file, [as in our Getting Started guide](https://developer.rackspace.com/docs/cloud-files/getting-started/?lang=go#upload-objects-to-container), this results in a file containing only "{}" being uploaded instead of the content that you actually provide!

I think we need a way to tell Perigee when the Content-Type is intentionally left blank.

smashwilson:
To trigger server-side content type detection, the Content-Type header should be *omitted* from the `PUT` request used to upload an object to Swift.

Fixes #336.

### DIFF
--- a/openstack/objectstorage/v1/objects/fixtures.go
+++ b/openstack/objectstorage/v1/objects/fixtures.go
@@ -105,13 +105,31 @@ func HandleListObjectNamesSuccessfully(t *testing.T) {
 	})
 }
 
-// HandleCreateObjectSuccessfully creates an HTTP handler at `/testContainer/testObject` on the test handler mux that
-// responds with a `Create` response.
-func HandleCreateObjectSuccessfully(t *testing.T) {
+// HandleCreateTextObjectSuccessfully creates an HTTP handler at `/testContainer/testObject` on the test handler mux
+// that responds with a `Create` response. A Content-Type of "text/plain" is expected.
+func HandleCreateTextObjectSuccessfully(t *testing.T) {
+	th.Mux.HandleFunc("/testContainer/testObject", func(w http.ResponseWriter, r *http.Request) {
+		th.TestMethod(t, r, "PUT")
+		th.TestHeader(t, r, "X-Auth-Token", fake.TokenID)
+		th.TestHeader(t, r, "Content-Type", "text/plain")
+		th.TestHeader(t, r, "Accept", "application/json")
+		w.WriteHeader(http.StatusCreated)
+	})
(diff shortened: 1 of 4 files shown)


openstack/objectstorage/v1/objects/fixtures.go
```diff
@@ -105,17 +105,35 @@ func HandleListObjectNamesSuccessfully(t *testing.T) {
...
+// HandleCreateTypelessObjectSuccessfully creates an HTTP handler at `/testContainer/testObject` on the test handler
+// mux that responds with a `Create` response. No Content-Type header may be present in the request, so that server-
+// side content-type detection will be triggered properly.
+func HandleCreateTypelessObjectSuccessfully(t *testing.T) {
+	th.Mux.HandleFunc("/testContainer/testObject", func(w http.ResponseWriter, r *http.Request) {
+		th.TestMethod(t, r, "PUT")
+		th.TestHeader(t, r, "X-Auth-Token", fake.TokenID)
+		th.TestHeader(t, r, "Accept", "application/json")
+
+		if contentType, present := r.Header["Content-Type"]; present {
```
smashwilson: I'm testing the `r.Header` map directly here so we fail even if a *blank* Content-Type header is set.

I don't know if Swift is that picky, but I figured it was better to be safe.

smashwilson:
@jrperritt This is ready for a review and (optimistically) a merge whenever you get the chance! :checkered_flag:

openstack/objectstorage/v1/objects/requests.go
```diff
@@ -205,14 +205,20 @@ func Create(c *gophercloud.ServiceClient, containerName, objectName string, cont
...
-	})
+	}
+
+	if contentType, explicit := h["Content-Type"]; explicit {
+		popts.ContentType = contentType
+		delete(h, "Content-Type")
+	} else {
+		popts.OmitContentType = true
+	}
+
```
jrperritt: :+1:

jrperritt:
:+1: :ship:

PR #343, opened 2015-01-27
Minor fixes to Cdn openstack rackspace

jrperritt:
These are just some minor changes to wrap up the CDN service PR.


smashwilson:
:+1:

PR #344, opened 2015-01-28
Improve CDN service patch updates.

smashwilson:
The CDN "services" resource accepts [PATCH updates in JSON Patch notation]((http://docs.rackspace.com/cdn/api/v1.0/cdn-devguide/content/PATCH_patchService__services__service_id__servicesOperations.html)). This will change the types to constrain what users can send to valid patches. I have an implementation stubbed out on the [Go playground](https://play.golang.org/p/b-TBVZIfuY), so this will apply it to Gophercloud proper.

 - [x] Bring in the Path type.
 - [x] Bring in the value interface.
  - [x] Implement value for Domains.
  - [x] Implement value for Origins.
  - [x] Implement value for CacheRules.
 - [x] Bring in the Patch interface.
  - [x] Addition
  - [x] Removal
  - [x] Replacement
 - [x] Accept a Patch array in Update.
 - [x] Get rid of the old UpdateOpts code.
 - [x] Extend Addition to support [either an exact index or "-" to append](https://tools.ietf.org/html/rfc6902#section-4.1).
 - [x] Extend the Patches to handle bulk operations.
 - [x] Update unit and acceptance tests to match.
 - [x] Support `/name` replacement.
 - [x] Support bulk removal.

### DIFF
--- a/openstack/cdn/v1/services/requests.go
+++ b/openstack/cdn/v1/services/requests.go
@@ -254,22 +254,6 @@ var (
 	PathCaching = Path{baseElement: "caching"}
 )
 
-// UpdateOpts represents the attributes used when updating an existing CDN service.
-type UpdateOpts []UpdateOpt
-
-// UpdateOpt represents a single update to an existing service. Multiple updates
-// to a service can be submitted at the same time. See UpdateOpts.
-type UpdateOpt struct {
-	// Specifies the update operation to perform.
-	Op Op `json:"op"`
-	// Specifies the JSON Pointer location within the service's JSON representation
-	// of the service parameter being added, replaced or removed.
-	Path string `json:"path"`
-	// Specifies the actual value to be added or replaced. It is not required for
-	// the remove operation.
-	Value map[string]interface{} `json:"value,omitempty"`
-}
-
 type value interface {
 	toPatchValue() map[string]interface{}
 	appropriatePath() Path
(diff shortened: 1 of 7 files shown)


smashwilson:
I dropped my sample code in [a gist](https://gist.github.com/smashwilson/1ad54e86b5a673677a60), too.

jrperritt:
Here is a gist with some modifications to the one above to allow `[]map[string]interface{}` and `map[string]interface{}` values for the `Value` field.
https://gist.github.com/jrperritt/5f75ea3667d3dcf514b0

smashwilson:
@jrperritt This is ready for review when you get a moment. I took your lead on bulk operations with a few modifications (basically I'm dispatching to another `Value` method rather than type-asserting, because type-asserting makes my OCD wince).

I currently support everything except:

 1. Operations on the `/name` path, and
 2. Bulk removal on a root path.

Do you think either of those are crucial enough to include in this PR? Both of them would require a new `Patch` implementation, basically.

openstack/cdn/v1/services/requests.go
```diff
@@ -209,75 +209,122 @@ func Get(c *gophercloud.ServiceClient, idOrURL string) GetResult {
...
+		"path":  r.Value.renderRootOr(func(p Path) string { return p.renderIndex(r.Index) }),
+		"value": r.Value.toPatchValue(),
+	}
+}
+
+// Removal is a Patch that requests the removal of a service parameter (Domain, Origin, or
+// CacheRule) by index. Pass it to the Update function as part of the Patch slice.
+type Removal struct {
+	Path  Path
+	Index int64
```
jrperritt: One idea to implement bulk removal would be to have an `All` field of type `bool` in the `Removal` struct, and then use the `renderRoot` method in the `ToCDNServiceUpdateMap` method if `All` is set to `true`.

openstack/cdn/v1/services/requests.go
```diff
@@ -209,75 +209,122 @@ func Get(c *gophercloud.ServiceClient, idOrURL string) GetResult {
...
+// Replacement is a Patch that alters a specific service parameter (Domain, Origin, or CacheRule)
+// in-place by index. Pass it to the Update function as part of the Patch slice.
+type Replacement struct {
+	Value value
+	Index int64
+}
+
+// ToCDNServiceUpdateMap converts a Replacement into a request body fragment suitable for the
+// Update call.
+func (r Replacement) ToCDNServiceUpdateMap() map[string]interface{} {
```
jrperritt: Yeah, implementing an update to `name` would be tricky here. Since it can only be a `replace` op, would it make sense to have a `Type` field in the `Patch` structs that had to be one of "name", "origins", "domains", ... ? If not for `flavor_id`, we'd just be able to check is the `value` is of type `string`.

jrperritt:
My initial thought is that users are going to want to be able to update the service name, and unfortunately PATCH is the only way to do it  :/

smashwilson:
> My initial thought is that users are going to want to be able to update the service name, and unfortunately PATCH is the only way to do it :/

No worries - I'll add it to the list. I think I'll just make a new `Patch` type for the name replacement operation, since it's so much of an outlier. Using an `All` field seems reasonable, too - a bit awkward because `Index` is ignored if it's set to `true`, but that already happens for bulk replacements.

jrperritt:
:+1:

smashwilson:
@jrperritt Got them in :sparkles:

That was less annoying than I thought it would be!

jrperritt:
This looks great to me. Much better than the previous PATCH implementation   :rocket:

PR #345, opened 2015-01-29
Add UpdateOpts type for consistency

jrperritt:
This PR adds an `UpdateOpts` type (of type `[]Patch`) to the `services` package to stay consistent with the other packages. It also update the unit and acceptance tests to reflect the addition.
There is also a fix for a minor bug in the `serviceasset` acceptance test; create a service was failing because one of the same name already existed (from the tests that creates a service in the `service` acceptance test).


smashwilson:
:metal:

PR #346, opened 2015-01-31
Add a missing omitempty on the CreateOpts struct of LBaaS members

haklop:
TenantId can only be set if the caller has an admin role.


jrperritt:
:+1:

PR #347, opened 2015-01-31
Add missing OK Codes for servers.Get

julienvey:
Without this OK Codes, for instance when a server is terminated and no longer available, no error is returned even when the request fails with a 404.


jrperritt:
:+1:

PR #348, opened 2015-02-01
Docs: Changing NewStorageV1 to NewObjectStorageV1

jackspirou:
Documentation Change

Changing `NewStorageV1` to `NewObjectStorageV1` for the object storage documentation page on gophercloud.io.

I was unable to find the method `NewStorageV1` as the current documentation suggests.  After scanning through the source code I found `NewObjectStorageV1` and I'm guessing that is the method the object storage documentation should reference.


jrperritt:
Quite right; good catch :)

PR #351, opened 2015-02-02
Add Update operation for Rackspace servers

jrperritt:


jamiehannaford:
Sorry to be a pain, but can we have some acceptance/unit tests :smile_cat: An extra fn on [L200](https://github.com/rackspace/gophercloud/blob/master/acceptance/rackspace/compute/v2/servers_test.go#L200) and [anywhere here](https://github.com/rackspace/gophercloud/blob/master/rackspace/compute/v2/servers/requests_test.go) would be awesome

jrperritt:
Sure :)

jrperritt:
There we go.

jrperritt:
Self-merging :shipit:

PR #352, opened 2015-02-02
Add IP v4 and v6 options to server CreateOpts

jrperritt:


jamiehannaford:
:+1: LGTM. I never knew these were legitimate options - are these documented anywhere?

jrperritt:
Not that I found, but they are in the source: https://github.com/openstack/nova/blob/master/nova/api/openstack/compute/servers.py#L528

PR #353, opened 2015-02-04
Make portID not mandatory in floatingips

julienvey:
With the current code, it is not possible to allocate a floatingIP without
assigning it to a port. Some tools will require to allocate an IP first to
assign it later.

This is allowed in [OpenStack Network API](http://developer.openstack.org/api-ref-networking-v2.html#layer3) 

This patch makes portID optional in gophercloud

### DIFF
--- a/openstack/networking/v2/extensions/layer3/floatingips/requests.go
+++ b/openstack/networking/v2/extensions/layer3/floatingips/requests.go
@@ -53,7 +53,6 @@ type CreateOpts struct {
 
 var (
 	errFloatingNetworkIDRequired = fmt.Errorf("A NetworkID is required")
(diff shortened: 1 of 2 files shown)


jrperritt:
To stay consistent with the other packages, can we use `omitempty` in the struct tag instead of a `*string`? I think we can remove [these lines](https://github.com/rackspace/gophercloud/blob/master/openstack/networking/v2/extensions/layer3/floatingips/requests.go#L91-94) and add the `omitempty` struct tag to make `PortID` optional, or is there more to accomplish than that?

julienvey:
Updated with your recommendation, thanks

jrperritt:
Thank you :) LGTM :+1:

PR #355, opened 2015-02-06
Fix openstack blockstorage attachments types

jtopjian:
This commit changes the blockstorage attachments type from
a slice of strings to a slice of string maps to better suit the
actual returned results.


jtopjian:
Closing and re-opening. I think it should be `map[string]interface{}` instead.

PR #357, opened 2015-02-06
OpenStack/Rackspace Orchestration

jrperritt:
This PR implements the Orchestration service for both OpenStack and Rackspace.

To Do:

- [x] OpenStack Stack Ops
- [x] Rackspace Stack Ops
- [x] OpenStack Stack Unit tests
- [x] Rackspace Stack Unit tests
- [x] OpenStack Stack Resources Ops
- [x] Rackspace Stack Resources Ops
- [x] OpenStack Stack Resources Unit tests
- [x] Rackspace Stack Resources Unit tests
- [x] OpenStack Stack Events Ops
- [x] Rackspace Stack Events Ops
- [x] OpenStack Stack Events Unit tests
- [x] Rackspace Stack Events Unit tests
- [x] OpenStack Templates Ops
- [x] Rackspace Templates Ops
- [x] OpenStack Templates Unit tests
- [x] Rackspace Templates Unit tests
- [x] OpenStack Build Info Ops
- [x] Rackspace Build Info Ops
- [x] OpenStack Build Info Unit tests
- [x] Rackspace Build Info Unit tests

### DIFF
--- a/openstack/orchestration/v1/apiversions/results.go
+++ b/openstack/orchestration/v1/apiversions/results.go
@@ -2,21 +2,16 @@ package apiversions
 
 import (
 	"github.com/mitchellh/mapstructure"
+	"github.com/rackspace/gophercloud"
 	"github.com/rackspace/gophercloud/pagination"
 )
-
-// Link represents the base URL for a particular version.
-type Link struct {
-	HRef string `mapstructure:"href"`
-	Rel  string `mapstructure:"rel"`
-}
 
 // APIVersion represents an API version for Neutron. It contains the status of
 // the API, and its unique ID.
 type APIVersion struct {
-	Status string `mapstructure:"status"`
-	ID     string `mapstructure:"id"`
-	Links  []Link `mapstructure:"links"`
+	Status string             `mapstructure:"status"`
+	ID     string             `mapstructure:"id"`
+	Links  []gophercloud.Link `mapstructure:"links"`
 }
 
 // APIVersionPage is the page returned by a pager when traversing over a
--- a/openstack/orchestration/v1/buildinfo/requests_test.go
+++ b/openstack/orchestration/v1/buildinfo/requests_test.go
@@ -1,0 +1,20 @@
+package buildinfo
+
+import (
+	"testing"
+
+	th "github.com/rackspace/gophercloud/testhelper"
+	fake "github.com/rackspace/gophercloud/testhelper/client"
+)
+
+func TestGetTemplate(t *testing.T) {
+	th.SetupHTTP()
+	defer th.TeardownHTTP()
+	HandleGetSuccessfully(t, GetOutput)
+
+	actual, err := Get(fake.ServiceClient()).Extract()
+	th.AssertNoErr(t, err)
+
+	expected := GetExpected
+	th.AssertDeepEquals(t, expected, actual)
+}
--- a/openstack/orchestration/v1/buildinfo/results.go
+++ b/openstack/orchestration/v1/buildinfo/results.go
@@ -1,0 +1,37 @@
+package buildinfo
+
+import (
+	"github.com/mitchellh/mapstructure"
+	"github.com/rackspace/gophercloud"
+)
+
+// Revision represents the API/Enging revision of a Heat deployment.
+type Revision struct {
+	Revision string `mapstructure:"revision"`
+}
+
+// BuildInfo represents the build information for a Heat deployment.
+type BuildInfo struct {
+	API    Revision `mapstructure:"api"`
+	Engine Revision `mapstructure:"engine"`
+}
+
+// GetResult represents the result of a Get operation.
+type GetResult struct {
+	gophercloud.Result
+}
+
+// Extract returns a pointer to a BuildInfo object and is called after a
+// Get operation.
+func (r GetResult) Extract() (*BuildInfo, error) {
+	if r.Err != nil {
+		return nil, r.Err
+	}
+
+	var res BuildInfo
+	if err := mapstructure.Decode(r.Body, &res); err != nil {
+		return nil, err
+	}
+
+	return &res, nil
+}
--- a/openstack/orchestration/v1/buildinfo/results.go
+++ b/openstack/orchestration/v1/buildinfo/results.go
@@ -5,7 +5,7 @@ import (
 	"github.com/rackspace/gophercloud"
 )
 
-// Revision represents the API/Enging revision of a Heat deployment.
+// Revision represents the API/Engine revision of a Heat deployment.
 type Revision struct {
 	Revision string `mapstructure:"revision"`
 }
--- a/openstack/orchestration/v1/stackevents/doc.go
+++ b/openstack/orchestration/v1/stackevents/doc.go
@@ -1,0 +1,3 @@
+// Package stackevents provides operations for finding, listing, and retrieving
+// stack events.
+package stackevents
--- a/openstack/orchestration/v1/stackevents/doc.go
+++ b/openstack/orchestration/v1/stackevents/doc.go
@@ -1,3 +1,4 @@
 // Package stackevents provides operations for finding, listing, and retrieving
-// stack events.
+// stack events. Stack events are events that take place on stacks such as
+// updating and abandoning.
 package stackevents
--- a/openstack/orchestration/v1/stackevents/requests.go
+++ b/openstack/orchestration/v1/stackevents/requests.go
@@ -1,0 +1,186 @@
+package stackevents
+
+import (
+	"github.com/racker/perigee"
+	"github.com/rackspace/gophercloud"
+	"github.com/rackspace/gophercloud/pagination"
+)
+
+// Find retreives stack events for the given stack name.
+func Find(c *gophercloud.ServiceClient, stackName string) FindResult {
+	var res FindResult
+
+	_, res.Err = perigee.Request("GET", findURL(c, stackName), perigee.Options{
+		MoreHeaders: c.AuthenticatedHeaders(),
+		Results:     &res.Body,
+		OkCodes:     []int{200},
+	})
+	return res
+}
+
+// SortDir is a type for specifying in which direction to sort a list of events.
+type SortDir string
+
+// SortKey is a type for specifying by which key to sort a list of events.
+type SortKey string
+
+// ResourceStatus is a type for specifying by which resource status to filter a
+// list of events.
+type ResourceStatus string
+
+// ResourceAction is a type for specifying by which resource action to filter a
+// list of events.
+type ResourceAction string
+
+var (
+	// SortAsc is used to sort a list of stacks in ascending order.
+	SortAsc SortDir = "asc"
+	// SortDesc is used to sort a list of stacks in descending order.
+	SortDesc SortDir = "desc"
+
+	// SortName is used to sort a list of stacks by name.
+	SortName SortKey = "name"
+	// SortResourceType is used to sort a list of stacks by resource type.
+	SortResourceType SortKey = "resource_type"
+	// SortCreatedAt is used to sort a list of stacks by date created.
+	SortCreatedAt SortKey = "created_at"
+)
+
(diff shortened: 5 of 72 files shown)


jrperritt:
@smashwilson @jamiehannaford This one's ready for review. It's quite a few LOC, but most of them are for unit tests.

openstack/orchestration/v1/apiversions/requests.go
```diff
@@ -0,0 +1,13 @@
+package apiversions
+
+import (
+	"github.com/rackspace/gophercloud"
+	"github.com/rackspace/gophercloud/pagination"
+)
+
+// ListVersions lists all the Neutron API versions available to end-users
+func ListVersions(c *gophercloud.ServiceClient) pagination.Pager {
```
jamiehannaford: For API version functionality, do you think it's better to shift it up a level to the `openstack/orchestration/apiversions` namespace?

openstack/orchestration/v1/apiversions/results.go
```diff
@@ -0,0 +1,47 @@
...
+import (
+	"github.com/mitchellh/mapstructure"
+	"github.com/rackspace/gophercloud/pagination"
+)
+
+// Link represents the base URL for a particular version.
+type Link struct {
+	HRef string `mapstructure:"href"`
+	Rel  string `mapstructure:"rel"`
+}
```
jamiehannaford: There's already a Link struct in root `gophercloud` (results.go)

openstack/orchestration/v1/buildinfo/requests_test.go
```diff
@@ -0,0 +1,20 @@
...
+func TestGetTemplate(t *testing.T) {
+	th.SetupHTTP()
+	defer th.TeardownHTTP()
+	HandleGetSuccessfully(t, GetOutput)
+
+	actual, err := Get(fake.ServiceClient()).Extract()
+	th.AssertNoErr(t, err)
+
+	expected := GetExpected
+	th.AssertDeepEquals(t, expected, actual)
```
jamiehannaford: how about `th.AssertDeepEquals(t, GetExpected, actual)`?

openstack/orchestration/v1/buildinfo/results.go
```diff
@@ -0,0 +1,37 @@
+package buildinfo
+
+import (
+	"github.com/mitchellh/mapstructure"
+	"github.com/rackspace/gophercloud"
+)
+
+// Revision represents the API/Enging revision of a Heat deployment.
```
jamiehannaford: s/Enging/Engine

openstack/orchestration/v1/stackevents/doc.go
```diff
@@ -0,0 +1,3 @@
+// Package stackevents provides operations for finding, listing, and retrieving
+// stack events.
+package stackevents
```
jamiehannaford: what is a stack event? :tongue:

openstack/orchestration/v1/stackevents/requests.go
```diff
@@ -0,0 +1,186 @@
+package stackevents
+
+import (
+	"github.com/racker/perigee"
+	"github.com/rackspace/gophercloud"
+	"github.com/rackspace/gophercloud/pagination"
+)
+
+// Find retreives stack events for the given stack name.
```
jamiehannaford: s/retreives/retrieves

openstack/orchestration/v1/stackevents/requests.go
```diff
@@ -0,0 +1,186 @@
...
+// the API. Marker and Limit are used for pagination.
+type ListOpts struct {
+	// The stack resource ID with which to start the listing.
+	Marker string `q:"marker"`
+	// Integer value for the limit of values to return.
+	Limit int `q:"limit"`
+	// Filters the event list by the specified ResourceAction. You can use this
+	// filter multiple times to filter by multiple resource actions: CREATE, DELETE,
+	// UPDATE, ROLLBACK, SUSPEND, RESUME or ADOPT.
+	ResourceActions []string `q:"resource_action"`
```
jamiehannaford: Are these resource actions locked? If so, maybe we could change it to `[]ResourceAction` and introduce each one as a var?

openstack/orchestration/v1/stackevents/requests.go
```diff
@@ -0,0 +1,186 @@
...
+	// Filters the event list by the specified resource_status. You can use this
+	// filter multiple times to filter by multiple resource statuses: IN_PROGRESS,
+	// COMPLETE or FAILED.
+	ResourceStatuses []string `q:"resource_status"`
```
jamiehannaford: Same here

openstack/orchestration/v1/stackevents/requests_test.go
```diff
@@ -0,0 +1,71 @@
...
+	th.CheckEquals(t, count, 1)
+}
+
+func TestListResourceEvents(t *testing.T) {
+	th.SetupHTTP()
+	defer th.TeardownHTTP()
+	HandleListResourceEventsSuccessfully(t, ListResourceEventsOutput)
+
+	count := 0
+	err := ListResourceEvents(fake.ServiceClient(), "hello_world", "49181cd6-169a-4130-9455-31185bbfc5bf", "my_resource", nil).EachPage(func(page pagination.Page) (bool, error) {
```
jamiehannaford: for the sake of brevity, could we re-use the same UUID as a const?

openstack/orchestration/v1/stackevents/results.go
```diff
@@ -0,0 +1,153 @@
...
+	"github.com/rackspace/gophercloud/pagination"
+)
+
+// Event represents a stack event.
+type Event struct {
+	ResourceName         string                 `mapstructure:"resource_name"`
+	Time                 time.Time              `mapstructure:"-"`
+	Links                []gophercloud.Link     `mapstructure:"links"`
+	LogicalResourceID    string                 `mapstructure:"logical_resource_id"`
+	ResourceStatusReason string                 `mapstructure:"resource_status_reason"`
```
jamiehannaford: could we have brief explanations for these fields?

openstack/orchestration/v1/stackresources/requests.go
```diff
@@ -0,0 +1,132 @@
+package stackresources
+
+import (
+	"github.com/racker/perigee"
+	"github.com/rackspace/gophercloud"
+	"github.com/rackspace/gophercloud/pagination"
+)
+
+// Find retreives stack resources for the given stack name.
```
jamiehannaford: s/retreives/retrieves

openstack/orchestration/v1/stacks/fixtures.go
```diff
@@ -0,0 +1,401 @@
...
+		th.TestHeader(t, r, "Accept", "application/json")
+
+		w.Header().Set("Content-Type", "application/json")
+		w.WriteHeader(http.StatusOK)
+		fmt.Fprintf(w, output)
+	})
+}
+
+/*
+`
```
jamiehannaford: can this be removed?

openstack/orchestration/v1/stacks/requests.go
```diff
@@ -0,0 +1,529 @@
...
+	// (OPTIONAL; REQUIRED IF TemplateURL IS EMPTY) A template to instantiate. The value
+	// is a stringified version of the JSON/YAML template. Since the template will likely
+	// be located in a file, one way to set this variable is by using ioutil.ReadFile:
+	// import "io/ioutil"
+	// var opts stacks.CreateOpts
+	// b, err := ioutil.ReadFile("path/to/you/template/file.json")
+	// if err != nil {
+	//   // handle error...
+	// }
+	// opts.Template = string(b)
```
jamiehannaford: :+1: great docs

openstack/orchestration/v1/stacks/requests_test.go
```diff
@@ -0,0 +1,231 @@
...
+	th.SetupHTTP()
+	defer th.TeardownHTTP()
+	HandleAbandonSuccessfully(t)
+
+	actual, err := Abandon(fake.ServiceClient(), "postman_stack", "16ef0584-4458-41eb-87c8-0dc8d5f66c87").Extract()
+	th.AssertNoErr(t, err)
+
+	expected := AbandonExpected
+	th.AssertDeepEquals(t, expected, actual)
+}
```
jamiehannaford: should this test be commented out?

jamiehannaford:
This is awesome - a lot of work :+1: Nearly all of my comments are minor issues. LGTM

openstack/orchestration/v1/apiversions/requests.go
jrperritt: Though I don't think it _would_ change, the API for `apiversions` could change in a later version of the Orchestration API. Also, we've already started doing it like this: `networking` and `blockstorage` have similiar `apiversions` packages.

openstack/orchestration/v1/stackevents/requests.go
jrperritt: Yeah, originally I had it like that. The issue I ran into was with the types in the `BuildQueryString` function. If we use `[]ResourceAction` then when we type switch in `BuildQueryString`, we'll have to have a switch statement for every new type like this.

openstack/orchestration/v1/stacks/fixtures.go
jrperritt: yes :)

openstack/orchestration/v1/stackevents/requests.go
jrperritt: I'm not sure which is better. What do you think @jamiehannaford ? @smashwilson Do you have any thoughts on this?

jamiehannaford: Hmm... how hard would it be to add a new fn that delegates mostly to `gophercloud.BuildQueryString` but also adds in support for `opts.ResourceActions` slice? it would just add to the query slice before calling `q.String()`, right?

jamiehannaford: I really don't mind, but I really like the idea of using strict types whenever we can because it inevitably helps the user IMO. But if it's not feasible in certain cases, generic types will do

openstack/orchestration/v1/apiversions/requests.go
jamiehannaford: okay, fine with me

openstack/orchestration/v1/stackevents/requests.go
smashwilson: Another idea would be to introduce an interface in `BuildQueryString` (`gophercloud.QueryStringable`?) and implement it for new types like `ResourceAction`. That gives us one thing that we can type-assert against in `BuildQueryString`, but also gives us the freedom to have stricter types.

jamiehannaford: Would each list operation need to implement their own query-string builder from scratch? Or would it be optional when custom types are used?

smashwilson: No, no - `ResourceAction` would implement `QueryStringable`, not `ListOpts`. `BuildQueryString` would still automatically handle strings and ints and so on automatically - you'd just need to implement it for non-basic types in option structs.

I *may* not be expressing myself terribly clearly this morning.

jamiehannaford: Ah, okay. So would `BuildQueryString` iterate over the `ListOpts` fields and detect whether a field implements `QueryStringable` or not? If it does, call the interface method; otherwise parse as usual.

smashwilson: Yeah, that's it exactly - basically, add `QueryStringable` to the list of type assertions that it attempts when it's iterating over fields.

jrperritt: OK, that's a good idea. I'll give that a shot.

jamiehannaford:
:rocket:

PR #359, opened 2015-02-07
Adds volumeattach extension to openstack compute v2

jtopjian:
This commit adds support for the `os-volume_attachments` nova v2 api extension.

### DIFF
--- a/acceptance/openstack/compute/v2/volumeattach_test.go
+++ b/acceptance/openstack/compute/v2/volumeattach_test.go
@@ -1,0 +1,127 @@
+// +build acceptance compute servers
+
+package v2
+
+import (
+	"os"
+	"testing"
+
+	"github.com/rackspace/gophercloud"
+	"github.com/rackspace/gophercloud/acceptance/tools"
+	"github.com/rackspace/gophercloud/openstack"
+	"github.com/rackspace/gophercloud/openstack/blockstorage/v1/volumes"
+	"github.com/rackspace/gophercloud/openstack/compute/v2/extensions/volumeattach"
+	"github.com/rackspace/gophercloud/openstack/compute/v2/servers"
+	th "github.com/rackspace/gophercloud/testhelper"
+)
+
+func newBlockClient(t *testing.T) (*gophercloud.ServiceClient, error) {
+	ao, err := openstack.AuthOptionsFromEnv()
+	th.AssertNoErr(t, err)
+
+	client, err := openstack.AuthenticatedClient(ao)
+	th.AssertNoErr(t, err)
+
+	return openstack.NewBlockStorageV1(client, gophercloud.EndpointOpts{
+		Region: os.Getenv("OS_REGION_NAME"),
+	})
+}
+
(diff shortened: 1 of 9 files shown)


openstack/compute/v2/extensions/volumeattach/requests.go
```diff
@@ -0,0 +1,109 @@
...
+
+	"github.com/racker/perigee"
+	"github.com/rackspace/gophercloud"
+	"github.com/rackspace/gophercloud/openstack/compute/v2/servers"
+	"github.com/rackspace/gophercloud/pagination"
+)
+
+// CreateOptsExt adds a VolumeAttachment option to the base CreateOpts.
+type CreateOptsExt struct {
+	servers.CreateOptsBuilder
```
jrperritt: `CreateOptsExt` is used to add extension parameters to the base `CreateOpts` from the `servers` package. After looking at the OpenStack Nova source code,  I couldn't find where a volume could be attached to a server at creation time. Were you able to get this to work for you?

jrperritt:
Looking good. Can you add a file with acceptance tests ([here](https://github.com/rackspace/gophercloud/tree/master/acceptance/openstack/compute/v2)) as well?

openstack/compute/v2/extensions/volumeattach/requests.go
jtopjian: You're right... I didn't actually check this.

jtopjian:
I'll give it a shot. I was hoping I could squeak by without it ;)

jtopjian:
How can I run the acceptance test for _just_ the file I create (and dependent utilities)

jrperritt:
You can run a specific file by passing part of the test function name to the `run` flag. For example, if the test function is named TestAttachVolume, you might run those tests like so:
```
go test -tags 'acceptance fixtures' -run "AttachVolume" -v github.com/rackspace/gophercloud/acceptance/openstack/compute/v2
```

jtopjian:
Thanks!

OK, got an acceptance test working.

acceptance/openstack/compute/v2/volumeattach_test.go
```diff
@@ -0,0 +1,127 @@
...
+
+	volume, err := createVAVolume(t, blockClient)
+	if err != nil {
+		t.Fatalf("Unable to create volume: %v", err)
+	}
+
+	createVolumeAttachment(t, computeClient, blockClient, server.ID, volume.ID)
+
+	defer func() {
+		servers.Delete(computeClient, server.ID)
```
jrperritt: You'll usually want to put `defer` statements immediately after the error checking lines for the same resource. The deferred function will be called even if the function errors out, but it has to called before then. For example, you'll want to move this function to delete the created server to line 104. Similarly for the function to delete the created volume.

jtopjian: OK, that makes sense.

jrperritt:
:+1: Great job, @jtopjian . Thank you for the PR and for your patience  :)

jtopjian:
Pleasure was all mine. Thank you for all your help!

PR #361, opened 2015-02-10
[rfr] Rackspace Cloud Databases and OpenStack Trove services

jamiehannaford:


openstack/db/v1/databases/requests.go
```diff
@@ -0,0 +1,117 @@
...
+
+// ToMap is a helper function to convert individual DB create opt structures
+// into sub-maps.
+func (opts CreateOpts) ToMap() (map[string]string, error) {
+	if opts.Name == "" {
+		return nil, fmt.Errorf("Name is a required field")
+	}
+	if len(opts.Name) > 64 {
+		return nil, fmt.Errorf("Name must be less than 64 chars long")
+	}
```
smashwilson: Is there a reason to enforce the character limit in the SDK, but not the constraints described in `CreateOpts`?

openstack/db/v1/databases/requests.go
```diff
@@ -0,0 +1,117 @@
...
+
+// ToDBCreateMap renders a JSON map for creating DBs.
+func (opts BatchCreateOpts) ToDBCreateMap() (map[string]interface{}, error) {
+	var dbs []map[string]string
+	for _, db := range opts {
+		dbMap, err := db.ToMap()
+		if err != nil {
+			return nil, err
+		}
+		dbs = append(dbs, dbMap)
```
smashwilson: We could also pre-allocate the slice of maps:

```go
dbs := make([]map[string]string, len(opts))
for i, db := range opts {
  dbMap, err := db.ToMap()
  if err != nil {
    return nil, err
  }
  dbs[i] = dbMap
}
```

It's a bit of a micro-optimization, but it saves the possibility of `append` unnecessarily re-allocating.

openstack/db/v1/databases/results.go
```diff
@@ -0,0 +1,72 @@
...
+	if err != nil {
+		return true, err
+	}
+	return len(dbs) == 0, nil
+}
+
+// NextPageURL will retrieve the next page URL.
+func (page DBPage) NextPageURL() (string, error) {
+	type resp struct {
+		Links []gophercloud.Link `mapstructure:"databases_links"`
```
smashwilson: I wish there was a cleaner way to handle this style of pagination, by `foo_links` element... :thought_balloon:

openstack/db/v1/flavors/results.go
```diff
@@ -0,0 +1,92 @@
...
+	ID string `mapstructure:"id"`
+
+	// The RAM capacity for the flavor.
+	RAM int `mapstructure:"ram"`
+
+	// The Name field provides a human-readable moniker for the flavor.
+	Name string `mapstructure:"name"`
+
+	// Links to access the flavor.
+	Links []gophercloud.Link
```
smashwilson: These links feel like a strange thing to expose. What are they useful for?

openstack/db/v1/instances/fixtures.go
```diff
@@ -0,0 +1,158 @@
...
+	resizeReq    = `{"resize": {"flavorRef": "2"}}`
+	resizeVolReq = `{"resize": {"volume": {"size": 4}}}`
+)
+
+var (
+	createResp        = fmt.Sprintf(`{"instance": %s}`, instance)
+	listInstancesResp = fmt.Sprintf(`{"instances":[%s]}`, instance)
+	getInstanceResp   = createResp
+	enableUserResp    = `{"user":{"name":"root","password":"secretsecret"}}`
+	isUserEnabledResp = `{"rootEnabled":true}`
```
smashwilson: Cool! Better than a bunch of copy-paste.

openstack/db/v1/instances/fixtures.go
```diff
@@ -0,0 +1,158 @@
...
+	Links: []gophercloud.Link{
+		gophercloud.Link{Href: "https://my-openstack.com/v1.0/1234/instances/1", Rel: "self"},
+	},
+	Name:   "json_rack_instance",
+	Status: "BUILD",
+	Volume: Volume{Size: 2},
+}
+
+func HandleCreate(t *testing.T) {
+	fixture.SetupHandler(t, rootURL, "POST", createReq, createResp, 200)
```
smashwilson: This is much cleaner too. :sparkles:

openstack/db/v1/instances/requests.go
```diff
@@ -0,0 +1,209 @@
...
+	// Specifies the volume size in gigabytes (GB). The value must be between 1
+	// and 300. Required.
+	Size int
+
+	// Name of the instance to create. The length of the name is limited to
+	// 255 characters and any characters are permitted. Optional.
+	Name string
+
+	// A slice of database information options.
+	Databases db.BatchCreateOpts
```
smashwilson: Do these need to be the batch creation types, or can we use the equivalent Builder interface types instead?

openstack/db/v1/instances/requests.go
```diff
@@ -0,0 +1,209 @@
...
+// ResizeInstance changes the memory size of the instance, assuming a valid
+// flavorRef is provided. It will also restart the MySQL service.
+func ResizeInstance(client *gophercloud.ServiceClient, id, flavorRef string) ActionResult {
+	var res ActionResult
+
+	reqBody := map[string]map[string]string{
+		"resize": map[string]string{
+			"flavorRef": flavorRef,
+		},
+	}
```
smashwilson: Hmm, this looks a bit awkward, especially with all of the `map[string]string[string]string[string]` stuff you need. How about this:

```go
type resize struct {
  FlavorRef string `json:"flavorRef"`
}

type req struct {
  Resize resize `json:"resize"`
}

reqBody := req{Resize: resize{FlavorRef: flavorRef}}
```

openstack/db/v1/instances/requests.go
```diff
@@ -0,0 +1,209 @@
...
+// only increasing the volume size and does not support decreasing the size.
+// The volume size is in gigabytes (GB) and must be an integer.
+func ResizeVolume(client *gophercloud.ServiceClient, id string, size int) ActionResult {
+	var res ActionResult
+
+	reqBody := map[string]map[string]map[string]int{
+		"resize": map[string]map[string]int{
+			"volume": map[string]int{"size": size},
+		},
+	}
```
smashwilson: Ditto here :point_up:

openstack/db/v1/instances/results.go
```diff
@@ -0,0 +1,156 @@
...
+	// The size in GB of the volume
+	Size int
+
+	Used float64
+}
+
+// Instance represents a remote MySQL instance.
+type Instance struct {
+	// Indicates the datetime that the instance was created
+	Created string //time.Time
```
smashwilson: Hmm. We do need to have a strategy for interpreting timestamp fields in responses.

This PR probably isn't the place, though.

openstack/db/v1/users/requests.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+
+	if opts.Host != "" {
+		user["host"] = opts.Host
+	}
+
+	var dbs []map[string]string
+	for _, db := range opts.Databases {
+		dbs = append(dbs, map[string]string{"name": db.Name})
+	}
```
smashwilson: Same as the other slice preallocation comment up above.

rackspace/db/v1/backups/requests.go
```diff
@@ -0,0 +1,138 @@
...
+)
+
+// CreateOptsBuilder is the top-level interface for creating JSON maps.
+type CreateOptsBuilder interface {
+	ToBackupCreateMap() (map[string]interface{}, error)
+}
+
+// CreateOpts is responsible for configuring newly provisioned backups.
+type CreateOpts struct {
+	// [REQUIRED] The name of the backup. The only restriction is the name must
```
smashwilson: I'd also love to have the required/optional annotations as struct tags, even if we don't take advantage of them just yet. It would at least standardize them across services!

rackspace/db/v1/configurations/requests.go
```diff
@@ -0,0 +1,287 @@
...
+
+// CreateOpts is the struct responsible for configuring new configurations.
+type CreateOpts struct {
+	// [REQUIRED] The configuration group name
+	Name string
+
+	// [REQUIRED] A map of user-defined configuration settings that will define
+	// how each associated datastore works. Each key/value pair is specific to a
+	// datastore type.
+	Values map[string]interface{}
```
smashwilson: Can we make this a `map[string]string`, or do the value types actually vary?

rackspace/db/v1/instances/delegate.go
```diff
@@ -0,0 +1,161 @@
...
+	// Specifies the volume size in gigabytes (GB). The value must be between 1
+	// and 300. Required.
+	Size int
+
+	// Name of the instance to create. The length of the name is limited to
+	// 255 characters and any characters are permitted. Optional.
+	Name string
+
+	// A slice of database information options.
+	Databases osDBs.BatchCreateOpts
```
smashwilson: We should probably use the interface types here, too.

smashwilson:
Nitpicking: complete. :checkered_flag: Looking good!

openstack/db/v1/databases/requests.go
jamiehannaford: I was unsure about this too - but I needed to draw the line somewhere. I'm not sure the SDK should handle complicated validation logic, but a simple character length check seemed simple enough because it was kind of related to the type itself. Then again, this seems like a tenuous link so I'm happy to remove this call if you think it's better to remove all validation

openstack/db/v1/databases/requests.go
jamiehannaford: :+1: good idea, will do

openstack/db/v1/databases/results.go
jamiehannaford: me too, I think it's something we can add to the roadmap

openstack/db/v1/flavors/results.go
jamiehannaford: yeah, IMHO they're not actually needed since a SDK should remove the need for canonical HTTP links. Is it safe to parse what's returned to user under the assumption that we think certain fields (like links) are not relevant/needed? Would other fields also fall into this pattern? cc/ @jrperritt

openstack/db/v1/instances/requests.go
jamiehannaford: how do you mean? so introduce another opts type for this sub-package and then delegate to the `db` package inside the builder function?

rackspace/db/v1/configurations/requests.go
jamiehannaford: the types will vary between strings and ints :/

jamiehannaford:
@smashwilson Thanks for the review - I took your advice and changed the pre-allocation slices and the awkward maps. The only things left for this PR, I think, are the `interface types` comments (which I need further explanation for) and whether we want to expose the links field :smile:

openstack/db/v1/flavors/results.go
jrperritt: I don't think there's anything wrong with exposing them. Users may want to store the resource's URL to use outside of Gophercloud.

smashwilson: They don't _hurt_, I just thought they looked odd. No action necessary :wink:

jrperritt: I agree; They're likely not to be used. Unfortunately, since we already have them in many of the other result structures, I can't get past my OCD to stop having them now ;)

jamiehannaford:
Update on this: Rackspace provider is complete and all integration tests pass. OpenStack provider is being more problematic - I've waiting on a few bug reports before integration tests can be completed.

openstack/db/v1/configurations/requests.go
```diff
@@ -0,0 +1,287 @@
...
+
+	if opts.Description != "" {
+		config["description"] = opts.Description
+	}
+
+	return map[string]interface{}{"configuration": config}, nil
+}
+
+// Create will create a new configuration group.
+func Create(client *gophercloud.ServiceClient, opts CreateOptsBuilder) CreateResult {
```
jrperritt: Do you have a link to the documentation you referenced for this PR? For example, I see this operation is the [Rackspace docs](http://api.rackspace.com/api-ref-databases.html#createConfiguration) but not in the [OpenStack docs](http://developer.openstack.org/api-ref-database-v1.html#configurations).

jrperritt: Ugh. The above was a bad example, because `create` actually does exist in the docs, but `get`, `update`, `delete`, `replace`, etc... don't.

jamiehannaford: Yeah, when I first implemented this _none_ of it existed in [these docs](http://developer.openstack.org/api-ref-database-v1.html), but now they do. The missing operations indicate either the docs are incomplete, or the DB implementation of this resource is incomplete. I don't really think it's the latter since I'm always finding missing methods in the docs... 

Since we're not testing this against a live API, I don't think it's so bad to make a few assumptions. Whenever we get a live API up and running (or somebody else reports a fault), we can refactor if necessary.

jamiehannaford:
@jrperritt Just a quick update: this is ready for review again after tweaking a few things I discovered when implementing into rack. Once this is merged, I'll create a separate PR on rack to vendor new changes, which will then be the base PR for the main chunk of CDB integration.

openstack/db/v1/configurations/results.go
```diff
@@ -0,0 +1,142 @@
...
+
+import (
+	"github.com/mitchellh/mapstructure"
+	"github.com/rackspace/gophercloud"
+	"github.com/rackspace/gophercloud/pagination"
+)
+
+// Config represents a configuration group API resource.
+type Config struct {
+	Created              string
```
jrperritt: Can we set these as `time.Time` and convert them in `Extract` similar to here: https://github.com/rackspace/gophercloud/blob/d783aa72ca0531c6d07f76c84fc73083a5e8af7f/rackspace/rackconnect/v3/cloudnetworks/results.go#L61

openstack/db/v1/databases/requests.go
```diff
@@ -0,0 +1,115 @@
...
+		db["character_set"] = opts.CharSet
+	}
+	if opts.Collate != "" {
+		db["collate"] = opts.Collate
+	}
+	return db, nil
+}
+
+// BatchCreateOpts allows for multiple databases to created and modified.
+type BatchCreateOpts []CreateOpts
```
jrperritt: I'm thinking this should be a `[]CreateOptsBuilder` and then have a `ToDBCreateMap` method on the `CreateOpts`, and then call the `ToDBCreateMap` method in the `BatchCreateOpts.ToDbCreateMap` for each `CreateOptsBuilder` in the slice.

openstack/db/v1/instances/requests.go
jrperritt: Similar to a comment above, instead of `db.BatchCreateOpts`, let's have the `[]db.CreateOptsBuilder`. I think that's what @smashwilson was proposing also.

openstack/db/v1/instances/requests.go
```diff
@@ -0,0 +1,238 @@
...
+
+	// Name of the instance to create. The length of the name is limited to
+	// 255 characters and any characters are permitted. Optional.
+	Name string
+
+	// A slice of database information options.
+	Databases db.BatchCreateOpts
+
+	// A slice of user information options.
+	Users users.BatchCreateOpts
```
jrperritt: Same as above :point_up:

openstack/db/v1/instances/requests.go
```diff
@@ -0,0 +1,238 @@
...
+		OkCodes:      []int{200},
+	})
+
+	return res.Body.(map[string]interface{})["rootEnabled"] == true, err
+}
+
+// RestartService will restart only the MySQL Instance. Restarting MySQL will
+// erase any dynamic configuration settings that you have made within MySQL.
+// The MySQL service will be unavailable until the instance restarts.
+func RestartService(client *gophercloud.ServiceClient, id string) ActionResult {
```
jrperritt: `Restart`?

openstack/db/v1/instances/requests.go
```diff
@@ -0,0 +1,238 @@
...
+		JSONBody: map[string]interface{}{"restart": struct{}{}},
+		OkCodes:  []int{202},
+	})
+
+	return res
+}
+
+// ResizeInstance changes the memory size of the instance, assuming a valid
+// flavorRef is provided. It will also restart the MySQL service.
+func ResizeInstance(client *gophercloud.ServiceClient, id, flavorRef string) ActionResult {
```
jrperritt: `Resize`?

openstack/db/v1/instances/results.go
jrperritt: Same as an above comment for converting to `time.Time` in `Extract`.

openstack/db/v1/users/requests.go
```diff
@@ -0,0 +1,132 @@
...
+
+	if len(dbs) > 0 {
+		user["databases"] = dbs
+	}
+
+	return user, nil
+}
+
+// BatchCreateOpts allows multiple users to be created at once.
+type BatchCreateOpts []CreateOpts
```
jrperritt: `[]CreateOptsBuilder`

rackspace/client.go
```diff
@@ -212,3 +212,13 @@ func NewRackConnectV3(client *gophercloud.ProviderClient, eo gophercloud.Endpoin
 	}
 	return &gophercloud.ServiceClient{ProviderClient: client, Endpoint: url}, nil
 }
+
+// NewOrchestrationV1 creates a ServiceClient that may be used to access the v1 DB service.
```
jrperritt: `NewDBV1`

rackspace/db/v1/backups/results.go
```diff
@@ -0,0 +1,94 @@
...
+type Backup struct {
+	Description string
+	ID          string
+	InstanceID  string `json:"instance_id" mapstructure:"instance_id"`
+	LocationRef string
+	Name        string
+	ParentID    string `json:"parent_id" mapstructure:"parent_id"`
+	Size        float64
+	Status      Status
+	Created     string
```
jrperritt: `time.Time`

rackspace/db/v1/instances/delegate.go
```diff
@@ -0,0 +1,49 @@
...
+// True if root user is enabled for the specified database instance or False
+// otherwise.
+func IsRootEnabled(client *gophercloud.ServiceClient, id string) (bool, error) {
+	return os.IsRootEnabled(client, id)
+}
+
+// RestartService will restart only the MySQL Instance. Restarting MySQL will
+// erase any dynamic configuration settings that you have made within MySQL.
+// The MySQL service will be unavailable until the instance restarts.
+func RestartService(client *gophercloud.ServiceClient, id string) os.ActionResult {
```
jrperritt: `Restart` (if we decide to change the OpenStack one similarly)

rackspace/db/v1/instances/delegate.go
```diff
@@ -0,0 +1,49 @@
...
+	return os.RestartService(client, id)
+}
+
+// ResizeInstance changes the memory size of the instance, assuming a valid
+// flavorRef is provided. It will also restart the MySQL service.
+func ResizeInstance(client *gophercloud.ServiceClient, id, flavorRef string) os.ActionResult {
```
jrperritt: Same as above :point_up:

rackspace/db/v1/instances/results.go
```diff
@@ -0,0 +1,135 @@
...
+	"github.com/rackspace/gophercloud/openstack/db/v1/datastores"
+	"github.com/rackspace/gophercloud/openstack/db/v1/flavors"
+	os "github.com/rackspace/gophercloud/openstack/db/v1/instances"
+	"github.com/rackspace/gophercloud/pagination"
+)
+
+// Instance represents a remote MySQL instance.
+type Instance struct {
+	// Indicates the datetime that the instance was created
+	Created string //time.Time
```
jrperritt: `time.Time`

jrperritt:
@jamiehannaford Gave it another read through. I think after the most recent comments are considered (not necessarily acted upon), this will be good to merge.

openstack/db/v1/instances/requests.go
jamiehannaford: @jrperritt `db.BatchCreateOpts` is the type that satisfies the [`db.CreateOptsBuilder` interface](https://github.com/jamiehannaford/gophercloud/blob/cloud-dbs/openstack/db/v1/databases/requests.go#L63) since it serialises each one and nests it in the JSON object. Do you want to get rid of the batch type? If so, what will be responsible for transforming the slice into a JSON object?

openstack/db/v1/databases/requests.go
jamiehannaford: @jrperritt I'm not sure how I'd get this to work. Most functions need to accept `[]CreateOpts` as an argument, and then call `ToDBCreateMap` to serialize it into a JSON object. 

If I instead made each `CreateOpts` satisfy `CreateOptsBuilder`, is it possible to define the `ToDBCreateMap` function for `[]CreateOptsBuilder` and make that slice satisfy the interface too? It says it's an invalid receiver. Defining it on the `BatchCreateOpts` alias works, but the consequence is that we need to specify a concrete type rather than an interface.

jrperritt: After looking at it again, I think this type is good as-is, and perhaps we change the type signatures that have `BatchCreateOpts` to `CreateOptsBuilder`.

jamiehannaford:
@jrperritt okay, I _think_ this is ready

jrperritt:
Looks great to me. Thank you @jamiehannaford for the hard work. :shipit: +2

PR #362, opened 2015-02-11
fix ListByServer URL and unit test

jrperritt:
This PR fixes the URL for the ListByServer function for the `secgroups` package.


jrperritt:
Merging :shipit:

PR #363, opened 2015-02-11
Fixed typo in image query requests

jtopjian:


jrperritt:
Well that would cause some headaches ;)
:+1:

Issue #354, opened 2015-02-05
Set a custom user agent


smashwilson:
This will help metrics collection, etc.

smashwilson:
Actually, this would dovetail nicely with the centralization of HTTP operations that will be needed for both #147 and #152.

It would also be beneficial to allow clients to easily append to the agent, to add specificity.

everett-toews:
First off http://tools.ietf.org/html/rfc7231#section-5.5.3

Not only do you want to set a custom user agent, you want to give your users the ability to _append_ to that custom user agent. So if your custom user agent is

    gophercloud/1.1.0

Provide a method for your users to easily append their own user agent so they can further customize it.

    gophercloud/1.1.0 docker-machine/0.1.2

This will give us much better insight into how people are using gophercloud. This needs to be part of *any * examples that are written for gophercloud so people just copy/paste it it.

Issue #147, opened 2014-03-06
Discussion: Support HTTP instrumentation, custom HTTP transports


sam-falvo:
This blog article has had an impact on my thinking: http://brandur.org/sdk .  I encourage everyone to read it.

In particular, the discussion surrounding HTTP instrumentation is directly relevant to Gophercloud.

Gophercloud presently relies on a package I wrote called Perigee, and while Perigee supports the installation of custom HTTP transports, Gophercloud itself doesn't expose this feature.  Finding the right way to expose this capability will, I think, be useful to have in the future.

Should I create a Perigee V3 API that exists purely as a Go interface and dependency-inject it somehow into Gophercloud?  Or should we rely on Go stdlib's HTTP transports?  There are arguments either way, and this should be an interesting topic of discussion once v0.2.x has been revealed.

jrperritt:
I believe this is closed by #365

PR #365, opened 2015-02-12
Centralize HTTP handling and allow custom http.Clients

smashwilson:
This is a prerequisite for a lot of must-have functionality: #354, #152, and #147, specifically.

In this pull request, I'll introduce the ProviderClient method and convert OpenStack blockstorage as an example. We can divvy up the rest of the service packages once this is merged (which will let us work them in parallel, for example). In the meantime, let me know what you think about the internal API - it's almost, but not quite, Perigee-like.

/cc @jrperritt @jamiehannaford

### DIFF
--- a/provider_client.go
+++ b/provider_client.go
@@ -1,5 +1,14 @@
 package gophercloud
 
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"net/http"
+)
+
 // ProviderClient stores details that are required to interact with any
 // services within a specific provider's API.
 //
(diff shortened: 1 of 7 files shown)


smashwilson:
Oh, yeah, I should do the pagination stuff, too.

smashwilson:
On second thought: I can automate the conversion fairly trivially. Working on a terrible, hacked-up Ruby script to do so presently.

provider_client.go
```diff
@@ -24,10 +33,155 @@ type ProviderClient struct {
...
+func (client *ProviderClient) Request(method, url string, options RequestOpts) (*http.Response, error) {
+	var body io.Reader
+	var contentType *string
+
+	// Derive the content body by either encoding an arbitrary object as JSON, or by taking a provided
+	// io.Reader as-is. Default the content-type to application/json.
+
+	if options.JSONBody != nil {
+		if options.RawBody != nil {
+			panic("Please provide only one of JSONBody or RawBody to gophercloud.Request().")
```
smashwilson: I'm calling `panic` here rather than return an error because providing both of these struct fields is a developer error, not a runtime error with the actual call.

provider_client.go
```diff
@@ -24,10 +33,155 @@ type ProviderClient struct {
...
+	// JSONResponse, if provided, will be populated with the contents of the response body parsed as
+	// JSON.
+	JSONResponse *interface{}
+	// OkCodes contains a list of numeric HTTP status codes that should be interpreted as success. If
+	// the response has a different code, an error will be returned.
+	OkCodes []int
+
+	// MoreHeaders specifies additional HTTP headers to be provide on the request. If a header is
+	// provided with a blank value (""), that header will be *omitted* instead: use this to suppress
+	// the default Accept header or an inferred Content-Type, for example.
```
smashwilson: I'm curious what the rest of you think about this as an alternative to the `OmitXyz` fields: if you provide a blank value for any header, it will be removed from the request. We gain some generality here.

On the other hand, it may be confusing at callsites, and there are only two headers that I'm currently inferring in `client.Request` anyway.

Any comments?

smashwilson: Another data point: both options are used exactly once :laughing:

jrperritt: :+1:  Fine by me. And if the header isn't present it will default?

smashwilson: The new default behavior for headers is:

 * If you provide a `JSONBody`, Content-Type defaults to `application/json`. If you provide a `RawBody` instead, or omit both, it defaults to omitted.
 * Accept defaults to `application/json` unless explicitly specified otherwise, or disabled (by passing `map[string]string{"Accept": ""}`).
 * `X-Auth-Token` is set on the request if a token is present on the client.

smashwilson:
Alright: I can automatically bulk-convert basically all of the remaining calls, but I'd rather do that in a follow-on PR (or PRs) to keep this one relatively legible. So, I believe this piece on its own is ready for review, now.

jrperritt:
This looks good to me. What do you think @jamiehannaford ?

jamiehannaford:
I really like it :+1: Do you think having `Get`, `Post`, `Delete` functions would be useful? I guess it's not that much different from specifying the method via a string. Maybe we might want to centralize method-specific functionality in the future?

smashwilson:
> I really like it  Do you think having Get, Post, Delete functions would be useful? I guess it's not that much different from specifying the method via a string. Maybe we might want to centralize method-specific functionality in the future?

Possibly. I was following Perigee's lead in this - it had method-specific functions but deprecated them in favor of a single `Request`.

We could always add them in later.

smashwilson:
Self-merging like a boss :shipit:

PR #366, opened 2015-02-12
Added the SecurityGroup attribute to servers result

jtopjian:


openstack/compute/v2/servers/results.go
```diff
@@ -139,6 +139,9 @@ type Server struct {
 	// AdminPass will generally be empty ("").  However, it will contain the administrative password chosen when provisioning a new server without a set AdminPass setting in the first place.
 	// Note that this is the ONLY time this field will be valid.
 	AdminPass string `json:"adminPass" mapstructure:"adminPass"`
+
+	// SecurityGroups includes the security groups that this instance has applied to it
+	SecurityGroups []map[string]interface{} `json:"security_groups" mapstructure:"security_groups"`
```
jrperritt: If each `SecurityGroups` element has the same format as a `secgroup.SecurityGroup`, we should convert them to that format.

jtopjian: AFAICT, It's only returning an array of names. No other attributes. Can you see otherwise if you do `nova --debug show instance`?

jrperritt: No, that's all I see also. OK, that's that then :)

PR #342, opened 2015-01-27
OpenStack Firewall as a Service

ggiamarchi:
__To-do list__

* Firewall rules
  * [x] CRUD implementation
  * [x] Unit tests
  * [x] Acceptance tests
* Firewall policies
  * [x] CRUD implementation
  * [x] Insert and remove rules implementation
  * [x] Unit tests
  * [x] Acceptance tests
* Firewalls
  * [x] CRUD implementation
  * [x] Unit tests
  * [x] Acceptance tests

PR #367, opened 2015-02-13
Use client.Request instead of perigee

smashwilson:
This is a follow-on PR to #365 that uses the new, centralized request method throughout the existing codebase. It uses a [terrible, hacked-up Ruby script](https://gist.github.com/smashwilson/aa1c19c0c4d679883f8a) to automate the conversion process (for all the easy cases).

### DIFF
--- a/openstack/cdn/v1/base/requests.go
+++ b/openstack/cdn/v1/base/requests.go
@@ -1,19 +1,14 @@
 package base
 
-import (
-	"github.com/rackspace/gophercloud"
-
-	"github.com/racker/perigee"
-)
+import "github.com/rackspace/gophercloud"
 
 // Get retrieves the home document, allowing the user to discover the
 // entire API.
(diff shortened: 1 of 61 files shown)


smashwilson:
```
ashl6947@MMN1F3FD58 ~/code/gophercloud (deperigee %=) $ git grep perigee
ashl6947@MMN1F3FD58 ~/code/gophercloud (deperigee %=) $ script/unittest >/dev/null && echo $?
0
```

smashwilson:
I'll keep the conversion script handy so I can use it again once #342 and #361 land. Thanks to the magic of `gofmt` we shouldn't have merge conflicts from this, though!

smashwilson:
Looks like we're green on Travis, too. This is ready for review, then! :sparkles:

jamiehannaford:
:shipit: :balloon:

PR #369, opened 2015-02-13
Custom User-Agent; Closes #354

### LINKED ISSUE #354: Set a custom user agent

jrperritt:
Adds UserAgent field and methods to ProviderClient

### DIFF
--- a/provider_client.go
+++ b/provider_client.go
@@ -159,6 +159,9 @@ func (client *ProviderClient) Request(method, url string, options RequestOpts) (
 		req.Header.Add(k, v)
 	}
 
+	// Set the User-Agent header
+	req.Header.Set("User-Agent", client.UserAgent.Join())
+
 	if options.MoreHeaders != nil {
 		for k, v := range options.MoreHeaders {
 			fmt.Printf("Applying header [%s: %v]\n", k, v)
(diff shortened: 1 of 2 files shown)


provider_client.go
```diff
@@ -146,15 +170,16 @@ func (client *ProviderClient) Request(method, url string, options RequestOpts) (
 		}
 	}
 
-	// Issue the request.
+	// Set user-agent header
+	req.Header.Set("User-Agent", client.UserAgent.Join())
```
smashwilson: I would do this *before* `options.MoreHeaders` is processed, up above. That lets `MoreHeaders` be "the final word" on what's actually sent in the request, which in turn keeps us from running into the "OmitAccept" kinds of scenarios down the road.

Not that I can foresee us omitting the user agent from a specific request for any reason - in this case it would be more about consistency.

provider_client.go
```diff
@@ -97,7 +124,6 @@ func (client *ProviderClient) Request(method, url string, options RequestOpts) (
 
 	// Derive the content body by either encoding an arbitrary object as JSON, or by taking a provided
 	// io.Reader as-is. Default the content-type to application/json.
-
```
smashwilson: Huh. `gofmt` discrepancies again? I shouldn't have had trailing whitespace... ah well.

jrperritt: Yeah, I don't know. All the comments had a blank line between them and the code on which they commented.

provider_client.go
jrperritt: Alrighty. Moved.

smashwilson:
:metal: :shipit:

PR #373, opened 2015-02-16
Auto reauth; Closes #152

### LINKED ISSUE #152: Handle automatic re-auth

jrperritt:


jrperritt:
Originally I added a `MaxRetries` field to the `AuthOptions` structure to make the number of re-authentication retries variable, but I don't know how helpful that'd be. If there is interest in adding it, it would be just a few lines. Currently it just tries to reauth once.

openstack/client.go
```diff
@@ -107,6 +108,10 @@ func v2auth(client *gophercloud.ProviderClient, endpoint string, options gopherc
 		return err
 	}
 
+	if options.AllowReauth {
+		client.ReauthFunc = AuthenticateV2
+		client.AuthOptions = options
+	}
```
smashwilson: You could also capture `AuthOptions` (and the client pointer) with a closure, instead:

```go
client.ReauthFunc = func() error {
  return AuthenticateV2(client, options)
}
```

Pros:
 * Cleaner (in my opinion, anyway).
 * Avoids unnecessarily exporting `AuthOptions` in the client struct.
 * Can capture more context in the future if we ever need it without breaking the function's type signature and therefore backwards compatibility (although we'd most likely need to add it to `AuthOptions` instead).

Cons:
 * A little more magical

provider_client.go
```diff
@@ -179,6 +189,19 @@ func (client *ProviderClient) Request(method, url string, options RequestOpts) (
 		return nil, err
 	}
 
+	if resp.StatusCode == 401 {
```
smashwilson: You could also use the constant here to avoid the magic number; `http.StatusUnauthorized`.

rackspace/client.go
```diff
@@ -49,6 +49,7 @@ func AuthenticatedClient(options gophercloud.AuthOptions) (*gophercloud.Provider
 	if err != nil {
 		return nil, err
 	}
+
```
smashwilson: Whitespace?

openstack/client.go
```diff
@@ -58,6 +58,7 @@ func AuthenticatedClient(options gophercloud.AuthOptions) (*gophercloud.Provider
 	if err != nil {
 		return nil, err
 	}
+
```
smashwilson: More whitespace?

smashwilson:
> Originally I added a MaxRetries field to the AuthOptions structure to make the number of re-authentication retries variable, but I don't know how helpful that'd be. If there is interest in adding it, it would be just a few lines. Currently it just tries to reauth once.

I like it fine the way it is :grin:. If reauthentication fails, the most likely cause is incorrect credentials (like a cycled API key), which retrying isn't going to help.

It might be worth detecting rate limiting and implementing a retry system with exponential backoff for that -- but that's a different status code and a different issue :wink:

openstack/client.go
jrperritt: :+1:  Good idea

smashwilson:
:metal:
ship it

PR #375, opened 2015-02-16
fixed issue with group id

longnguyen11288:
it need to be group_id to be correctly parsed by nova


jrperritt:
Aha! Great catch. :+1:

PR #376, opened 2015-02-17
Use client.Request instead of perigee in FWaaS

ggiamarchi:


jrperritt:
Excellent. Thank you :)

PR #377, opened 2015-02-18
Get All Pages; Closes #298

### LINKED ISSUE #298: Pagination: Get all the things
Proposed Addition to pager.Pagination:

```go

import (
	identityRoles "github.com/rackspace/gophercloud/rackspace/identity/v2/roles"
	identAdminRoles "github.com/rackspace/gophercloud/openstack/identity/v2/extensions/admin/roles"
)

pager := identityRoles.List(sc)

p, err := pager.AllPages()
if err != nil {
   return err
}

roles, err := identAdminRoles.ExtractRoles(p)
if err != nil {
   return err
}

println("Got em all")
```

jrperritt:
This PR adds an `AllPages` method to allow users to retrieve all pages from a `List` operation at once. When complete, it will close #298 .
As implemented, each `List` function will need to add an additional line to set the `PageType` in the `Pager` structure. Currently, I've implemented it in 3 packages, each corresponding to a different page type: containers (marker), volumes (single), and servers (linked). If this implementation is acceptable, I'll change the other `List` operations similarly.

### DIFF
--- a/pagination/pager.go
+++ b/pagination/pager.go
@@ -140,8 +140,8 @@ func (p Pager) AllPages() (Page, error) {
 		return nil, err
 	}
 
-	// Switch on the page body type. Recognized types are `map[string]interface{}`
-	// and `[]byte`.
+	// Switch on the page body type. Recognized types are `map[string]interface{}`,
+	// `[]byte`, and `[]interface{}`.
 	switch testPage.GetBody().(type) {
 	case map[string]interface{}:
 		// key is the map key for the page body if the body type is `map[string]interface{}`.
@@ -196,6 +196,7 @@ func (p Pager) AllPages() (Page, error) {
 		if err != nil {
 			return nil, err
 		}
+		// Set body to value of type `[]interface{}`
 		body = reflect.MakeSlice(reflect.TypeOf(pagesSlice), len(pagesSlice), len(pagesSlice))
 		for i, s := range pagesSlice {
 			body.Index(i).Set(reflect.ValueOf(s))
--- a/pagination/pager.go
+++ b/pagination/pager.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"net/http"
 	"reflect"
+	"strings"
 
 	"github.com/rackspace/gophercloud"
 )
@@ -154,7 +155,7 @@ func (p Pager) AllPages() (Page, error) {
 			b := page.GetBody().(map[string]interface{})
 			for k := range b {
 				// If it's a linked page, we don't want the `links`, we want the other one.
-				if k != "links" {
+				if !strings.HasSuffix(k, "links") {
 					key = k
 				}
 			}
--- a/pagination/pager.go
+++ b/pagination/pager.go
@@ -3,6 +3,7 @@ package pagination
 import (
 	"errors"
 	"fmt"
+	"net/http"
 	"reflect"
 
 	"github.com/rackspace/gophercloud"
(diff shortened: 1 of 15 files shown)


jrperritt:
Er, the tests are passing. One Go version got hung up in TravisCI.

pagination/pager.go
```diff
@@ -113,3 +124,106 @@ func (p Pager) EachPage(handler func(Page) (bool, error)) error {
...
+	switch testPage.GetBody().(type) {
+	case map[string]interface{}:
+		// key is the map key for the page body if the body type is `map[string]interface{}`.
+		var key string
+		// Iterate over the pages to concatenate the bodies.
+		err := p.EachPage(func(page Page) (bool, error) {
+			b := page.GetBody().(map[string]interface{})
+			for k := range b {
+				// If it's a linked page, we don't want the `links`, we want the other one.
+				if k != "links" {
```
jamiehannaford: This is basically trying to extract all the elements out of a nested object without knowing the resource key, right? The only thing we know is that we want to ignore links. But most OpenStack link structures are often in the form of `servers_links` not `links`. So maybe we should use `strings.Contains` here instead?

pagination/pager.go
```diff
@@ -113,3 +124,106 @@ func (p Pager) EachPage(handler func(Page) (bool, error)) error {
...
+		if err != nil {
+			return nil, err
+		}
+		// Set body to value of type `map[string]interface{}`
+		body = reflect.MakeMap(reflect.MapOf(reflect.TypeOf(key), reflect.TypeOf(pagesSlice)))
+		body.SetMapIndex(reflect.ValueOf(key), reflect.ValueOf(pagesSlice))
+	case []byte:
+		// Iterate over the pages to concatenate the bodies.
+		err := p.EachPage(func(page Page) (bool, error) {
+			b := page.GetBody().([]byte)
```
jamiehannaford: what about `page.GetBody().([]byte) + []byte{10}`?

pagination/pager.go
```diff
@@ -113,3 +124,106 @@ func (p Pager) EachPage(handler func(Page) (bool, error)) error {
...
+	default:
+		return nil, fmt.Errorf("Page body has unrecognized type.")
+	}
+
+	// Each `Extract*` function is expecting a specific type of page coming back,
+	// otherwise the type assertion in those functions will fail. PageType is needed
+	// to create a type in this method that has the same type that the `Extract*`
+	// function is expecting and set the Body of that object to the concatenated
+	// pages.
+	page := reflect.New(reflect.TypeOf(p.PageType))
```
jamiehannaford: I have no idea whether this is possible, but doesn't the current implementation *already* hint at what the specific Page type is? For example, the page fn for servers is:

```go
createPageFn := func(r pagination.PageResult) pagination.Page {
		return ServerPage{pagination.LinkedPageBase{PageResult: r}}
}
```

Could we capture this `ServerPage` type in the switch statement (using some kind of reflect magic) and then use here? We wouldn't need to introduce a new Pager field then...

pagination/pager.go
jrperritt: :+1: I used `strings.HasSuffix` which should be a bit more robust.

pagination/pager.go
jrperritt: Hmm, quite right. I'd forgotten about that. That makes it much nicer. Changed.

jamiehannaford:
:rocket: looks great

PR #381, opened 2015-02-22
os-networks extension

jtopjian:
This commit adds the os-networks extention. This can be used to view
details about the nova-network-based networks that a tenant has access
to.

### DIFF
--- a/acceptance/openstack/compute/v2/network_test.go
+++ b/acceptance/openstack/compute/v2/network_test.go
@@ -1,0 +1,78 @@
+// +build acceptance compute servers
+
+package v2
+
+import (
+	"os"
+	"testing"
+
+	"github.com/rackspace/gophercloud"
+	"github.com/rackspace/gophercloud/openstack/compute/v2/extensions/networks"
+	"github.com/rackspace/gophercloud/openstack/compute/v2/servers"
+	th "github.com/rackspace/gophercloud/testhelper"
+)
+
+func getNetworkIDFromNetworkExtension(t *testing.T, client *gophercloud.ServiceClient, networkName string) (string, error) {
+	allPages, err := networks.List(client).AllPages()
+	if err != nil {
+		t.Fatalf("Unable to list networks: %v", err)
+	}
+
+	networkList, err := networks.ExtractNetworks(allPages)
+	if err != nil {
+		t.Fatalf("Unable to list networks: %v", err)
+	}
+
+	networkID := ""
+	for _, network := range networkList {
+		t.Logf("Network: %v", network)
+		if network.Label == networkName {
+			networkID = network.ID
+		}
+	}
+
+	t.Logf("Found network ID for %s: %s\n", networkName, networkID)
+
+	return networkID, nil
+}
+
+func TestNetworks(t *testing.T) {
+	networkName := os.Getenv("OS_NETWORK_NAME")
+	if networkName == "" {
+		t.Fatalf("OS_NETWORK_NAME must be set")
+	}
+
+	choices, err := ComputeChoicesFromEnv()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	client, err := newClient()
+	if err != nil {
+		t.Fatalf("Unable to create a compute client: %v", err)
+	}
+
+	networkID, err := getNetworkIDFromNetworkExtension(t, client, networkName)
+	if err != nil {
+		t.Fatalf("Unable to get network ID: %v", err)
+	}
+
+	// createNetworkServer is defined in tenantnetworks_test.go
+	server, err := createNetworkServer(t, client, choices, networkID)
+	if err != nil {
+		t.Fatalf("Unable to create server: %v", err)
+	}
+	defer func() {
+		servers.Delete(client, server.ID)
+		t.Logf("Server deleted.")
+	}()
+
+	if err = waitForStatus(client, server, "ACTIVE"); err != nil {
+		t.Fatalf("Unable to wait for server: %v", err)
+	}
+
+	allPages, err := networks.List(client).AllPages()
+	allNetworks, err := networks.ExtractNetworks(allPages)
+	th.AssertNoErr(t, err)
+	t.Logf("Retrieved all %d networks: %+v", len(allNetworks), allNetworks)
+}
--- a/openstack/compute/v2/extensions/networks/fixtures.go
+++ b/openstack/compute/v2/extensions/networks/fixtures.go
@@ -1,0 +1,209 @@
+// +build fixtures
+
+package networks
+
+import (
+	"fmt"
+	"net/http"
+	"testing"
+	"time"
+
+	th "github.com/rackspace/gophercloud/testhelper"
+	"github.com/rackspace/gophercloud/testhelper/client"
+)
+
+// ListOutput is a sample response to a List call.
+const ListOutput = `
+{
+    "networks": [
+        {
+            "bridge": "br100",
+            "bridge_interface": "eth0",
+            "broadcast": "10.0.0.7",
+            "cidr": "10.0.0.0/29",
+            "cidr_v6": null,
+            "created_at": "2011-08-15 06:19:19.387525",
+            "deleted": false,
+            "deleted_at": null,
+            "dhcp_start": "10.0.0.3",
+            "dns1": null,
+            "dns2": null,
+            "gateway": "10.0.0.1",
+            "gateway_v6": null,
+            "host": "nsokolov-desktop",
+            "id": "20c8acc0-f747-4d71-a389-46d078ebf047",
+            "injected": false,
+            "label": "mynet_0",
+            "multi_host": false,
+            "netmask": "255.255.255.248",
+            "netmask_v6": null,
+            "priority": null,
+            "project_id": "1234",
+            "rxtx_base": null,
+            "updated_at": "2011-08-16 09:26:13.048257",
+            "vlan": 100,
+            "vpn_private_address": "10.0.0.2",
+            "vpn_public_address": "127.0.0.1",
+            "vpn_public_port": 1000
+        },
+        {
+            "bridge": "br101",
+            "bridge_interface": "eth0",
+            "broadcast": "10.0.0.15",
+            "cidr": "10.0.0.10/29",
+            "cidr_v6": null,
+            "created_at": "2011-08-15 06:19:19.885495",
+            "deleted": false,
+            "deleted_at": null,
+            "dhcp_start": "10.0.0.11",
+            "dns1": null,
+            "dns2": null,
+            "gateway": "10.0.0.9",
+            "gateway_v6": null,
+            "host": null,
+            "id": "20c8acc0-f747-4d71-a389-46d078ebf000",
+            "injected": false,
+            "label": "mynet_1",
+            "multi_host": false,
+            "netmask": "255.255.255.248",
+            "netmask_v6": null,
+            "priority": null,
+            "project_id": null,
+            "rxtx_base": null,
+            "updated_at": null,
+            "vlan": 101,
+            "vpn_private_address": "10.0.0.10",
+            "vpn_public_address": null,
+            "vpn_public_port": 1001
+        }
+    ]
+}
+`
+
+// GetOutput is a sample response to a Get call.
+const GetOutput = `
+{
+    "network": {
+			"bridge": "br101",
+			"bridge_interface": "eth0",
+			"broadcast": "10.0.0.15",
+			"cidr": "10.0.0.10/29",
+			"cidr_v6": null,
+			"created_at": "2011-08-15 06:19:19.885495",
+			"deleted": false,
+			"deleted_at": null,
+			"dhcp_start": "10.0.0.11",
+			"dns1": null,
+			"dns2": null,
+			"gateway": "10.0.0.9",
+			"gateway_v6": null,
+			"host": null,
+			"id": "20c8acc0-f747-4d71-a389-46d078ebf000",
+			"injected": false,
+			"label": "mynet_1",
+			"multi_host": false,
+			"netmask": "255.255.255.248",
+			"netmask_v6": null,
+			"priority": null,
+			"project_id": null,
+			"rxtx_base": null,
+			"updated_at": null,
+			"vlan": 101,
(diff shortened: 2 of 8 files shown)


jtopjian:
This is only implementing List and Get. There are several more actions described [here](http://developer.openstack.org/api-ref-compute-v2-ext.html#ext-os-networks), but in my experience, these are both admin-only actions and they are usually done when the cloud is first built. Network management is not as flexible as with Neutron.

jrperritt:
Looks like you have a `.swp` file in here  :)

acceptance/openstack/compute/v2/network_test.go
```diff
@@ -0,0 +1,104 @@
...
+	return server, err
+}
+
+func TestNetwork(t *testing.T) {
+	networkName := os.Getenv("OS_NETWORK_NAME")
+	if networkName == "" {
+		t.Fatalf("OS_NETWORK_NAME must be set")
+	}
+
+	choices, err := ComputeChoicesFromEnv()
```
jrperritt: If you move `OS_NETWORK_NAME` to `ComputeChoices`, you can put the `networkName` check after this and check for `choices.NetworkName`.

openstack/compute/v2/extensions/networks/fixtures.go
```diff
@@ -0,0 +1,207 @@
...
+            "vpn_public_port": 1001
+        }
+    ]
+}
+`
+
+// GetOutput is a sample response to a Get call.
+const GetOutput = `
+{
+    "network": {
```
jrperritt: Looks like the formatting got away here :)

openstack/compute/v2/extensions/networks/fixtures.go
```diff
@@ -0,0 +1,207 @@
...
+	ProjectID:         "1234",
+	RXTXBase:          "",
+	UpdatedAt:         "2011-08-16 09:26:13.048257",
+	VLAN:              "100",
+	VPNPrivateAddress: "10.0.0.2",
+	VPNPublicAddress:  "127.0.0.1",
+	VPNPublicPort:     "1000",
+}
+
+// SecondNetwork is the first result in ListOutput.
```
jrperritt: s/first/second

openstack/compute/v2/extensions/networks/fixtures.go
```diff
@@ -0,0 +1,207 @@
...
+		th.TestMethod(t, r, "GET")
+		th.TestHeader(t, r, "X-Auth-Token", client.TokenID)
+
+		w.Header().Add("Content-Type", "application/json")
+		fmt.Fprintf(w, ListOutput)
+	})
+}
+
+// HandleGetSuccessfully configures the test server to respond to a Get request
+// for an existing network
```
jrperritt: s/network/network.

openstack/compute/v2/extensions/networks/results.go
```diff
@@ -0,0 +1,143 @@
...
+
+	// VPNPrivateAddress is the private address of the CloudPipe VPN.
+	VPNPrivateAddress string `mapstructure:"vpn_private_address"`
+
+	// VPNPublicAddress is the public address of the CloudPipe VPN.
+	VPNPublicAddress string `mapstructure:"vpn_public_address"`
+
+	// VPNPublicPort is the port of the CloudPipe VPN.
+	VPNPublicPort string `mapstructure:"vpn_public_port"`
+}
```
jrperritt: We'll want to parse `DeletedAt`, `CreatedAt`, and `UpdatedAt` to `time.Time` objects similar to [here](https://github.com/rackspace/gophercloud/blob/master/openstack/orchestration/v1/stacks/results.go#L89). Notice the `mapstructure` struct tags for these fields is "-".

openstack/compute/v2/extensions/networks/results.go
```diff
@@ -0,0 +1,143 @@
...
+	VLAN string `mapstructure:"vlan"`
+
+	// VPNPrivateAddress is the private address of the CloudPipe VPN.
+	VPNPrivateAddress string `mapstructure:"vpn_private_address"`
+
+	// VPNPublicAddress is the public address of the CloudPipe VPN.
+	VPNPublicAddress string `mapstructure:"vpn_public_address"`
+
+	// VPNPublicPort is the port of the CloudPipe VPN.
+	VPNPublicPort string `mapstructure:"vpn_public_port"`
```
jrperritt: Can this be anything more than an `int`? I can't find documentation on these fields.

jrperritt:
Looking good. Just a few changes to make.

openstack/compute/v2/extensions/networks/results.go
jtopjian: Same - I went by the unquoted numbers in the API examples. Logically it can't be anything other than `int` if it's a port number. 

Should I change it to `string` for safety? Or is there another type that's better?

jtopjian:
Awesome. Thank you for the review.

openstack/compute/v2/extensions/networks/results.go
jrperritt: Well it's a `string` right now. If it can be a range of ports, then it needs to be a `string`. If it's just one number, then it should be an `int`.

jtopjian: Oh - duh. Sorry, when I was writing this out, I had in my head that it should be `int` but looks like I made it `string` to be safe. Port range is not possible here, so I'll go ahead and make it int.

Also, `vpn_public_port` is `int` in the actual db schema, so I think that solves it. :)

Now that I'm looking at the db schema, `vlan`, `priority`, and `rxtx_base` are also `int`s, so I'll change those as well.

jrperritt: No worries, I knew what you meant  :)

:+1:

jtopjian:
OK, made some changes. I left it as a separate commit for now. I'll squash everything at the end.

AllPages was failing on the changes I had to make for the Extract functions:

```
=== RUN TestNetwork
--- FAIL: TestNetwork (0.23s)
panic: interface conversion: interface is map[string][]interface {}, not map[string]interface {} [recovered]
        panic: interface conversion: interface is map[string][]interface {}, not map[string]interface {}

goroutine 5 [running]:
testing.func·006()
        /usr/local/go/src/testing/testing.go:441 +0x181
github.com/rackspace/gophercloud/openstack/compute/v2/extensions/networks.ExtractNetworks(0x7fe438f12688, 0xc2081e2870, 0x0, 0x0, 0x0, 0x0, 0x0)
        /root/go/src/github.com/rackspace/gophercloud/openstack/compute/v2/extensions/networks/results.go:119 +0x187
github.com/rackspace/gophercloud/acceptance/openstack/compute/v2.getNetworkID(0xc2080521b0, 0xc208213320, 0x7fffb0b45e3e, 0x6, 0x0, 0x0, 0x0, 0x0)
        /root/go/src/github.com/rackspace/gophercloud/acceptance/openstack/compute/v2/network_test.go:22 +0x1de
github.com/rackspace/gophercloud/acceptance/openstack/compute/v2.TestNetwork(0xc2080521b0)
        /root/go/src/github.com/rackspace/gophercloud/acceptance/openstack/compute/v2/network_test.go:86 +0x2eb

```

Not sure if that's something I did or a bug with AllPages?

jtopjian:
Oh, also: I'll make the acc test changes wrt Choices after that gets merged.

openstack/compute/v2/extensions/networks/results.go
```diff
@@ -0,0 +1,212 @@
...
+
+// ExtractNetworks interprets a page of results as a slice of Networks
+func ExtractNetworks(page pagination.Page) ([]Network, error) {
+	var res struct {
+		Networks []Network `mapstructure:"networks"`
+	}
+
+	err := mapstructure.WeakDecode(page.(NetworkPage).Body, &res)
+
+	rawNetworks := (((page.(NetworkPage).Body).(map[string]interface{}))["networks"]).([]interface{})
```
jrperritt: This is where a run-time error is getting thrown. `page.(NetworkPage).Body` is of type `map[string][]interface{}` (not `map[string]interface{}`. Changing that type assertion should get it to work. Let me know if doesn't.

jtopjian: Yup - tried that and it gave me a compile error off the bat.

jrperritt: Oh, I see the problem. The `List` object is a `map[string]interface{}` and the `AllPages` object is a `map[string][]interface{}`. Hmm, I'll think of something.

jrperritt: See jtopjian/os-networks#1 for a possible fix.

openstack/compute/v2/extensions/networks/results.go
```diff
@@ -0,0 +1,200 @@
...
+// ExtractNetworks interprets a page of results as a slice of Networks
+func ExtractNetworks(page pagination.Page) ([]Network, error) {
+	var res struct {
+		Networks []Network `mapstructure:"networks"`
+	}
+
+	config := &mapstructure.DecoderConfig{
+		Result: &res,
+		DecodeHook: func(from reflect.Kind, to reflect.Kind, data interface{}) (interface{}, error) {
+			if from == reflect.String && to == reflect.Struct {
```
jtopjian: I'm lost. This part isn't getting hit at all. I changed `String` to `Map` and was eventually able to parse the time, but the changes weren't being reflected in `res` for the return.

jrperritt:
Ok. I _think_ I have a fix for the problem. When #384 gets merged, pull in the changes and revert back to [this commit](https://github.com/jtopjian/gophercloud/commit/6c7e892616f4fa8bfa1d66a92f705a1f76872e70) . After doing that, all the tests are passing for me.

jtopjian:
OK, that sounds good -- but I think we're getting tangled up.

To clarify, the bug that I was running into was when using `AllPages` such as here in the first commit of this PR:

https://github.com/jtopjian/gophercloud/blob/5c39ac87e749d23ba7d4a4444a7da594672314dd/acceptance/openstack/compute/v2/network_test.go#L16-L37

But after I made the `time.Time` changes.

The second commit (6c7e892616f) reverted back to paging rather than `AllPages` to get the tests to pass. So technically, reverting back to 6c7e892616f will pass no matter what, since it's not using `AllPages`.

The commit you made included an `AllPages` test at the end of of the file which would also trigger this issue.

So, if I understand everything, once #384 gets merged, I should be able to revert to 6c7e892616f since that includes the time parsing, re-add the `AllPages` test from the first commit, and also add your `AllPages` test from your commit and everything _should_ work? :smiley:

jrperritt:
Actually, after some more testing, there's more that needs to be done :/  I'll have to think about it some more.

jrperritt:
If you need the `os-networks` operations immediately, then you can just type assert in the `Extract*` function to see if the page body is `map[string]interface{}` or `map[string][]interface` until a more elegant solution is thought up and implemented.

jtopjian:
Yeah, that's the direction I was taking last night:

```go
if _, ok := body.(map[string][]interface{}); ok {
```

Is there a better way of checking?

jtopjian:
Forgot to mention: either way, I'm happy to wait until a proper solution is in place rather than hack stuff that will need changed later. :smile:

jrperritt:
You can do check the type that way or with a type switch:
```
switch page.(NetworkPage).Body.(type) {
case map[string]interface{}:
    // do some things
case map[string][]interface{}:
    // do some other things
default:
    // return an error about an unknown type
}
```
Type switch might be cleaner, but it's up to you.

jrperritt:
@jtopjian Unless we decide to change packages or get a PR accepted to the `mapstructure` package, type switching (or asserting) may be the best (read: only) way to go. It's only necessary when there's a `time.Time` in the result structure. The `mapstructure` package requires a `map` type when converting to a `struct` (such as `time.Time`), but a `time.Time` structure doesn't have any exported fields, so it can't be created from a `map`.

jtopjian:
OK, Understood.

Here's my latest attempt. I feel it's sloppy and there's a more efficient way to do this, but I've been fiddling with it for too long.

And how's this for dark irony: the date fields (and some other fields) are _only_ retrieved when an admin issues the command, not a normal user. Those fields will just have their nil value.

It looks like there is an undocumented `os-tenant-networks` endpoint that retrieves all of the user-accessible fields in a more concise result. From playing around with the `nova` cli tool, calling the various `net-*` and `network-*` sub-commands use a mix of both `os-networks` and `os-tenant-networks`, so it's not as if this work was all for naught :)

Anyway, let me know if anything can be cleaned up.

jtopjian:
@jrperritt After thinking about it for a day or so, I think this should be abandoned in favor of `os-tenant-networks`. While there isn't proper documentation on it, the Nova code defines it very clearly. I'll have a PR for that shortly.

Thoughts?

I owe you a beer for the trouble with this PR.

jrperritt:
It's up to you. Do `os-tenant-networks` and `nova-networks` share the same functionality?

Ha, on the contrary, not only was this PR not troublesome, it shed light on an issue that I need to resolve in some other packages, so for that I thank you  :)

jtopjian:
`os-tenant-networks` looks like a watered-down version of `os-networks` that allows a user to obtain some basic information of the networks they have access to. It doesn't include everything, such as the v6 subnet. The create functionality of `os-tenant-networks` looks very strange as it requires very little information.

`os-networks` provides an exhaustive list of information to the admin and a lesser amount (but still more than `os-tenant-networks`) to a regular user.

Neither of my PRs include support for creating. If they're merged as-is, I'll probably eventually get that finished just for completeness, but I need to set up a special nova-network environment to do that. For now, my main drive for these two PRs is to allow nova-network users to get some basic information on existing networks.

Feel free to take both :smile:

jtopjian:
@jrperritt Hi there!

I'm getting around to cleaning up these old PRs. I rebased, cleaned a few things up, and squashed as an attempt at a clean slate review. Do you mind looking it over when you get a chance?

jrperritt:
Howdy, @jtopjian . Great, I should have some time on Friday to review it.

jrperritt:
LGTM. +2

jtopjian:
Thank you, sir!

PR #382, opened 2015-02-22
Adds os-server-groups

jtopjian:
This commit adds support for the os-server-groups extension. This allows
users to create scheduling policies for launching instances.

### DIFF
--- a/openstack/compute/v2/extensions/servergroups/fixtures.go
+++ b/openstack/compute/v2/extensions/servergroups/fixtures.go
@@ -1,0 +1,161 @@
+// +build fixtures
+
+package servergroups
+
+import (
+	"fmt"
+	"net/http"
+	"testing"
+
+	th "github.com/rackspace/gophercloud/testhelper"
+	"github.com/rackspace/gophercloud/testhelper/client"
+)
+
+// ListOutput is a sample response to a List call.
+const ListOutput = `
+{
+    "server_groups": [
+        {
+            "id": "616fb98f-46ca-475e-917e-2563e5a8cd19",
+            "name": "test",
+            "policies": [
+                "anti-affinity"
+            ],
+            "members": [],
+            "metadata": {}
+        },
+        {
+            "id": "4d8c3732-a248-40ed-bebc-539a6ffd25c0",
+            "name": "test2",
+            "policies": [
+                "affinity"
+            ],
+            "members": [],
+            "metadata": {}
+        }
+    ]
+}
+`
+
+// GetOutput is a sample response to a Get call.
+const GetOutput = `
+{
+    "server_group": {
+        "id": "616fb98f-46ca-475e-917e-2563e5a8cd19",
+        "name": "test",
+        "policies": [
+            "anti-affinity"
+        ],
+        "members": [],
+        "metadata": {}
+    }
+}
+`
+
+// CreateOutput is a sample response to a Post call
+const CreateOutput = `
+{
+    "server_group": {
+        "id": "616fb98f-46ca-475e-917e-2563e5a8cd19",
+        "name": "test",
+        "policies": [
+            "anti-affinity"
+        ],
+        "members": [],
+        "metadata": {}
+    }
+}
+`
+
+// FirstServerGroup is the first result in ListOutput.
+var FirstServerGroup = ServerGroup{
+	ID:   "616fb98f-46ca-475e-917e-2563e5a8cd19",
+	Name: "test",
+	Policies: []string{
+		"anti-affinity",
+	},
+	Members:  []string{},
+	Metadata: map[string]interface{}{},
+}
+
+// SecondServerGroup is the second result in ListOutput.
+var SecondServerGroup = ServerGroup{
+	ID:   "4d8c3732-a248-40ed-bebc-539a6ffd25c0",
+	Name: "test2",
+	Policies: []string{
+		"affinity",
+	},
+	Members:  []string{},
+	Metadata: map[string]interface{}{},
+}
+
+// ExpectedServerGroupSlice is the slice of results that should be parsed
+// from ListOutput, in the expected order.
+var ExpectedServerGroupSlice = []ServerGroup{FirstServerGroup, SecondServerGroup}
+
+// CreatedServerGroup is the parsed result from CreateOutput.
+var CreatedServerGroup = ServerGroup{
+	ID:   "616fb98f-46ca-475e-917e-2563e5a8cd19",
+	Name: "test",
+	Policies: []string{
+		"anti-affinity",
+	},
+	Members:  []string{},
+	Metadata: map[string]interface{}{},
+}
+
+// HandleListSuccessfully configures the test server to respond to a List request.
+func HandleListSuccessfully(t *testing.T) {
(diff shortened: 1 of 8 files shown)


jtopjian:
The scheduler hints extension will need added for instances to take advantage of this. I'll try working on that next.

jtopjian:
Any update on this? This PR will create fully functional server groups, but the caveat is that they aren't actually usable until instances support scheduling hints. Would it be preferable to get that work done prior to this being merged in?

jrperritt:
Sorry for the delay; I didn't know if this was ready for review.

If it's not too much trouble, I think that's the best course of action. That is, let's get dependencies merged first.

jtopjian:
It looks like I caught myself in a chicken/egg scenario.

There are [a lot](http://docs.openstack.org/juno/config-reference/content/section_compute-scheduler.html) of filters available, but the only ones enabled by default that can take advantage of scheduling hints are server groups.

So in order to write a proper acceptance test of scheduling hints, server groups needs to be available :smile: 

I have all of the scheduling hint work done, but to cleanly merge everything, it'd be best if this PR was merged first. Once that's done, I can open a new PR with the appropriate changes.

Thoughts?

jrperritt:
OK, I'll try to find some time this week to review, though I'm not sure I'll be able to test it.

jtopjian:
Sounds good. 

I'd be surprised if you couldn't test it. Server Groups are a default feature beginning with Icehouse. [Here's](http://dev.cloudwatt.com/en/blog/affinity-and-anti-affinity-in-openstack.html) a good writeup. Let me know if you run into issues, though.

jtopjian:
Alternatively, I could just submit the scheduling / server group functionality in one PR.

openstack/compute/v2/extensions/servergroups/fixtures.go
```diff
@@ -0,0 +1,161 @@
...
+	ID:   "616fb98f-46ca-475e-917e-2563e5a8cd19",
+	Name: "test",
+	Policies: []string{
+		"anti-affinity",
+	},
+	Members:  []string{},
+	Metadata: map[string]interface{}{},
+}
+
+// SecondServerGroup is the first result in ListOutput.
```
jrperritt: s/first/second

openstack/compute/v2/extensions/servergroups/requests.go
```diff
@@ -0,0 +1,84 @@
...
+)
+
+// List returns a Pager that allows you to iterate over a collection of ServerGroups.
+func List(client *gophercloud.ServiceClient) pagination.Pager {
+	return pagination.NewPager(client, listURL(client), func(r pagination.PageResult) pagination.Page {
+		return ServerGroupsPage{pagination.SinglePageBase(r)}
+	})
+}
+
+// CreateOptsBuilder describes struct types that can be accepted by the Create call. Notable, the
```
jrperritt: s/Notable/Notably

openstack/compute/v2/extensions/servergroups/requests.go
```diff
@@ -0,0 +1,84 @@
...
+	Policies []string
+}
+
+// ToServerGroupCreateMap constructs a request body from CreateOpts.
+func (opts CreateOpts) ToServerGroupCreateMap() (map[string]interface{}, error) {
+	if opts.Name == "" {
+		return nil, errors.New("Missing field required for server group creation: Name")
+	}
+
+	if len(opts.Policies) < 1 {
```
jrperritt: :+1:

openstack/compute/v2/extensions/servergroups/requests.go
```diff
@@ -0,0 +1,84 @@
...
+func Get(client *gophercloud.ServiceClient, id string) GetResult {
+	var res GetResult
+	_, res.Err = client.Request("GET", getURL(client, id), gophercloud.RequestOpts{
+		JSONResponse: &res.Body,
+		OkCodes:      []int{200},
+	})
+	return res
+}
+
+// Delete requests the deletion of a previous allocated ServerGroup.
```
jrperritt: s/previous/previously

openstack/compute/v2/extensions/servergroups/requests.go
```diff
@@ -0,0 +1,84 @@
...
+func Delete(client *gophercloud.ServiceClient, id string) DeleteResult {
+	var res DeleteResult
+	_, res.Err = client.Request("DELETE", deleteURL(client, id), gophercloud.RequestOpts{
+		OkCodes: []int{202},
```
jrperritt: I think this should include a `204`, which actually segues nicely into a new feature that we have. Now the `ProviderClient` has all the HTTP verbs as methods, in addition to default `OkCodes`. So, for example, this would be equivalent to what's there now: `_, res.Err = client.Delete(deleteURL)`. We're going to be using that format going forward. Of course, if the `OkCodes` are not the common ones for a given HTTP verb, they can be passed in as they are now. Reference: https://github.com/rackspace/gophercloud/blob/master/provider_client.go#L294

openstack/compute/v2/extensions/servergroups/results.go
```diff
@@ -0,0 +1,87 @@
...
+
+import (
+	"github.com/mitchellh/mapstructure"
+	"github.com/rackspace/gophercloud"
+	"github.com/rackspace/gophercloud/pagination"
+)
+
+// A ServerGroup creates a policy for instance placement in the cloud
+type ServerGroup struct {
+	// ID is a unique ID of the Server Group.
```
jrperritt: s/a/the

openstack/compute/v2/extensions/servergroups/results.go
```diff
@@ -0,0 +1,87 @@
...
+
+	return response.ServerGroups, err
+}
+
+type ServerGroupResult struct {
+	gophercloud.Result
+}
+
+// Extract is a method that attempts to interpret any Server Group resource
+// response as a Server Group struct.
```
jrperritt: s/Server Group/ServerGroup

jrperritt:
Looks really good. Mostly insignificant grammatical changes (such is my way). The only code-related change suggested is the use of the new HTTP verb methods, which I don't think were available to you when you first made this PR  :)

jtopjian:
Awesome - thank you the review. I consider the comment wording to be just as important since it turns into documentation, so no worries about that.

I think I made the appropriate verb changes. Everything is still working, so that's always promising :)

openstack/compute/v2/extensions/servergroups/requests.go
```diff
@@ -0,0 +1,77 @@
...
+func Get(client *gophercloud.ServiceClient, id string) GetResult {
+	var res GetResult
+	_, res.Err = client.Get(getURL(client, id), &res.Body, nil)
+	return res
+}
+
+// Delete requests the deletion of a previously allocated ServerGroup.
+func Delete(client *gophercloud.ServiceClient, id string) DeleteResult {
+	var res DeleteResult
+	_, res.Err = client.Delete(deleteURL(client, id), nil)
```
jrperritt: :+1:

jrperritt:
LGTM :rocket:

PR #384, opened 2015-02-25
[wip] Bug fix in AllPages method

jrperritt:
This PR fixes a bug that was encountered while coding PR #381 . It converts whatever the page body type to an `interface{}` to be consistent with `gophercloud.Response.Body`. A conflict arises when, for example, the `List` operation returns a `map[string]interface{}` and the `AllPages` method returns a `map[string][]interface{}` and there is type assertion logic in the `Extract*` method.


jrperritt:
After some more testing, this needs some work done to it.

PR #386, opened 2015-02-27
os-tenant-networks

jtopjian:
This commit adds support for the os-tenant-networks extension.

### DIFF
--- a/openstack/compute/v2/extensions/tenantnetworks/doc.go
+++ b/openstack/compute/v2/extensions/tenantnetworks/doc.go
@@ -1,0 +1,2 @@
+// Package tenantnetworks provides the ability for tenants to see information about the networks they have access to
+package tenantnetworks
--- a/openstack/compute/v2/extensions/tenantnetworks/results.go
+++ b/openstack/compute/v2/extensions/tenantnetworks/results.go
@@ -1,0 +1,68 @@
+package tenantnetworks
+
+import (
+	"github.com/mitchellh/mapstructure"
+	"github.com/rackspace/gophercloud"
+	"github.com/rackspace/gophercloud/pagination"
+)
+
+// A Network represents a nova-network that an instance communicates on
+type Network struct {
+	// CIDR is the IPv4 subnet.
+	CIDR string `mapstructure:"cidr"`
+
+	// ID is the UUID of the network.
+	ID string `mapstructure:"id"`
+
+	// Name is the common name that the network has.
+	Name string `mapstructure:"label"`
+}
+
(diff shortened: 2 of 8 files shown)


jtopjian:
While investigating #381, I came across the [os-tenant-networks](https://github.com/openstack/nova/blob/master/nova/api/openstack/compute/contrib/os_tenant_networks.py) extension. This looks to be a better solution to the problem I was trying to solve with #381 which is the ability for a tenant to retrieve basic information about the networks they have access to. 

This extension works with both nova-network and Neutron, though it's more beneficial to nova-network users. Neutron does not return the CIDR field, but could still be used as an alternative/easier way of mapping networks to UUIDs. 

This commit implements the List and Get actions, but not the Create action. I could implement the Create action, but I'll need a special environment to do this. Just like with #381, I don't anticipate Create being used often (or at all) with this extension.

jtopjian:
Also, I'm not sure why the build is failing. Running the tests locally are passing. The logs seem to point to something with the networking/v2 API? Maybe I'm reading it wrong.

openstack/compute/v2/extensions/tenantnetworks/doc.go
```diff
@@ -0,0 +1,2 @@
+// Package tenantnetwork provides the ability for tenants to see information about the networks they have access to
```
jrperritt: /tenantnetwork/tenantnetworks

openstack/compute/v2/extensions/tenantnetworks/results.go
```diff
@@ -0,0 +1,68 @@
...
+
+// A Network represents a nova-network that an instance communicates on
+type Network struct {
+	// CIDR is the IPv4 subnet.
+	CIDR string `mapstructure:"cidr"`
+
+	// ID is the UUID of the network.
+	ID string `mapstructure:"id"`
+
+	// Label is the common name that the network has..
```
jrperritt: One too many periods.

openstack/compute/v2/extensions/tenantnetworks/results.go
```diff
@@ -0,0 +1,68 @@
...
+	Label string `mapstructure:"label"`
```
jrperritt: Can we call this `Name` for consistency?

jrperritt:
Looks good. Some very minor changes from me. I'm not worried about the build failure in Travis; it looks unrelated to this.

jtopjian:
Cool - thanks for the review. Please see latest version.

jrperritt:
Just one more

jrperritt:
:+1:

PR #392, opened 2015-03-05
Unit test failing on Go Tip

jrperritt:
This PR adds a filler line to avoid a runtime issue

### DIFF
--- a/openstack/networking/v2/extensions/external/requests.go
+++ b/openstack/networking/v2/extensions/external/requests.go
@@ -1,6 +1,10 @@
 package external
 
-import "github.com/rackspace/gophercloud/openstack/networking/v2/networks"
+import (
+	"time"
+
+	"github.com/rackspace/gophercloud/openstack/networking/v2/networks"
+)
 
 // AdminState gives users a solid type to work with for create and update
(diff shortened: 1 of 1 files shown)


openstack/networking/v2/extensions/external/requests.go
```diff
@@ -25,6 +29,14 @@ type CreateOpts struct {
 
 // ToNetworkCreateMap casts a CreateOpts struct to a map.
 func (o CreateOpts) ToNetworkCreateMap() (map[string]interface{}, error) {
+
+	// DO NOT REMOVE. Though this line seemingly does nothing of value, it is a
+	// splint to prevent the unit test from failing on Go Tip. We suspect it is a
+	// compiler issue that will hopefully be worked out prior to our next release.
+	// Again, for all the unit tests to pass, this line is necessary and sufficient
+	// at the moment.
```
smashwilson: Want to add a note about the current Go version so we don't forget? "Reassess after go 1.5 is release" or something?

smashwilson: */me bikesheds on a do-nothing, comment-only PR*

jrperritt: Added  :)

PR #393, opened 2015-03-05
Add logic to handle AllPages result page; Closes #389

### LINKED ISSUE #389: Using AllPages method with `Extract*s` functions that have dates causes run-time error
The body of a response is an `interface{}`, but the body of an `AllPages` response is an `[]interface`. The `mapstructure` package handles this fine, but type switches need to be added to the `Extract*s` functions that use dates, since `mapstructure` can't decode from `string` to `time.Time`.

jrperritt:
PR to resolve #389

### DIFF
--- a/openstack/orchestration/v1/stackevents/results.go
+++ b/openstack/orchestration/v1/stackevents/results.go
@@ -1,6 +1,8 @@
 package stackevents
 
 import (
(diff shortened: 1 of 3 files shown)


jrperritt:
Merging because bug :shipit:

PR #394, opened 2015-03-06
fix recursive reauth failure

kazegusuri:
If both password and token are specified, token is always used in Identity API v2.
After token was expired, the expired token is used for re-authentication. This fixes recursive re-authetication failure by clearing token before calling reauth method.

### DIFF
--- a/openstack/client.go
+++ b/openstack/client.go
@@ -109,6 +109,7 @@ func v2auth(client *gophercloud.ProviderClient, endpoint string, options gopherc
 
 	if options.AllowReauth {
 		client.ReauthFunc = func() error {
+			client.TokenID = ""
 			return AuthenticateV2(client, options)
 		}
 	}


jrperritt:
Though there's nothing wrong with zeroing-out the token before re-authenticating, I don't think the expired token is used even if provided as a header to a `Create` token call. A new token is automatically generated [here] (https://github.com/rackspace/gophercloud/blob/master/openstack/client.go#L98) and uses the `Username` and `Password` provided in `gophercloud.AuthOptions` for the initial authentication.

kazegusuri:
I see, I will check it again next week.

kazegusuri:
After some investigations, I found `tokens2.Create` always set `X-Auth-Token` header via  [here](https://github.com/rackspace/gophercloud/blob/master/provider_client.go#L163) if provided TokenID (which means TokenID is non-empty). This is why re-authentication fails.
There can be option not to set  `X-Auth-Token` header at `tokens2.Create`, but clearing TokenID is easy IMO. How do you think?

jrperritt:
Hmm, which cloud provider are you using? We can merge this in, I'm just curious if different providers handle this differently.

kazegusuri:
Thank you for merging. I use OpenStack Icehouse for now.

PR #395, opened 2015-03-09
[wip] Rackspace Neutron Security Groups; Closes #391

### LINKED ISSUE #391: Neutron Security Groups (Rackspace)

jrperritt:
This PR implements Neutron Security Groups for Rackspace. When merged, it will close #391

### DIFF
--- a/rackspace/networking/v2/ports/delegate.go
+++ b/rackspace/networking/v2/ports/delegate.go
@@ -24,6 +24,9 @@ func Get(c *gophercloud.ServiceClient, networkID string) os.GetResult {
 
 // Create accepts a CreateOpts struct and creates a new network using the values
 // provided. You must remember to provide a NetworkID value.
+//
+// NOTE: Currently the SecurityGroup option is not implemented to work with
+// Rackspace.
 func Create(c *gophercloud.ServiceClient, opts os.CreateOptsBuilder) os.CreateResult {
 	return os.Create(c, opts)
 }
--- a/rackspace/networking/v2/security/doc.go
+++ b/rackspace/networking/v2/security/doc.go
@@ -1,0 +1,32 @@
+// Package security contains functionality to work with security group and
+// security group rules Neutron resources.
+//
+// Security groups and security group rules allows administrators and tenants
+// the ability to specify the type of traffic and direction (ingress/egress)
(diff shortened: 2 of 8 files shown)


rackspace/networking/v2/security/doc.go
```diff
@@ -0,0 +1,32 @@
...
+//
+// The basic characteristics of Neutron Security Groups are:
+//
+// For ingress traffic (to an instance)
+//  - Only traffic matched with security group rules are allowed.
+//  - When there is no rule defined, all traffic are dropped.
+//
+// For egress traffic (from an instance)
+//  - Only traffic matched with security group rules are allowed.
+//  - When there is no rule defined, all egress traffic are dropped.
```
jamiehannaford: should it be "traffic are dropped" or "traffic is dropped"?

jamiehannaford:
:+1: LGTM

rackspace/networking/v2/ports/delegate.go
```diff
@@ -24,6 +24,9 @@ func Get(c *gophercloud.ServiceClient, networkID string) os.GetResult {
 
 // Create accepts a CreateOpts struct and creates a new network using the values
 // provided. You must remember to provide a NetworkID value.
+//
+// NOTE: Currently the SecurityGroup option is not implemented to work with
+// Rackspace.
```
smashwilson: I wish we had a way to log a warning if users tried to use it, for example. Logging from a library seems like a bad idea, though.

smashwilson:
Looks reasonable to me!

rackspace/networking/v2/security/doc.go
jrperritt: Haha, I'm not sure; they're your words ;)  I think it should be 'is'. I've changed it there and in the OpenStack Neutron `security` doc.go file.

rackspace/networking/v2/ports/delegate.go
jrperritt: Yeah, I know, not great. This is another reason we need a completely well-posed schema from which to build the API.

jrperritt:
Thank you for the review @smashwilson and @jamiehannaford   :smile:

PR #396, opened 2015-03-10
NetworkExtAttrs bug fix

jrperritt:
This PR weakly decodes the output from NetworkExt operations to correctly convert types (such as `float64` to `string`).


jrperritt:
Merging bug fix :shipit:

PR #399, opened 2015-03-18
RackConnect v3; Closes #364

### LINKED ISSUE #364: Rackconnect support
Ref: http://www.managaedcolocation.net/knowledge_center/article/getting-started-with-the-rackconnect-v30-api

Are there plans to add rackconnect support? I am happy to help.

jrperritt:
This PR add support for RackConnect v3. When merged, it'll close #364


rackspace/rackconnect/v3/cloudnetworks/requests.go
```diff
@@ -0,0 +1,27 @@
...
+	return pagination.NewPager(c, url, createPage)
+}
+
+// Get retrieves a specific cloud network (that is associated with RackConnect)
+// based on its unique ID.
+func Get(c *gophercloud.ServiceClient, id string) GetResult {
+	var res GetResult
+	_, res.Err = c.Request("GET", getURL(c, id), gophercloud.RequestOpts{
+		JSONResponse: &res.Body,
+		OkCodes:      []int{200},
```
jamiehannaford: idea (unrelated from this PR): could we add default `OkCodes` to the base `Request` func? Most GETs return 200, most POSTs return 201, etc. So we could provide a default and allow overriding where necessary - could shave a few dozen lines out of the codebase

rackspace/rackconnect/v3/lbpools/doc.go
```diff
@@ -0,0 +1,14 @@
+// Package lbpools provides access to load balancer pools associated with a
+// RackConnect configuration. Load Balancer Pools must be configured in advance
+// by your Network Security team to be elegible for use with RackConnect.
```
jamiehannaford: s/elegible/eligible/

rackspace/rackconnect/v3/lbpools/requests.go
```diff
@@ -0,0 +1,156 @@
...
+	var res GetNodeDetailsResult
+	_, res.Err = c.Request("GET", nodeDetailsURL(c, poolID, nodeID), gophercloud.RequestOpts{
+		JSONResponse: &res.Body,
+		OkCodes:      []int{200},
+	})
+	return res
+}
+
+// NodesOpts are options for bulk adding/deleting nodes to LB pools.
+type NodesOpts struct {
```
jamiehannaford: should this be `NodeOpts` since it's a singular construction? you then pass in multiple as a slice

rackspace/rackconnect/v3/lbpools/requests.go
```diff
@@ -0,0 +1,156 @@
...
+	ServerID string
+	PoolID   string
+}
+
+// CreateNodes adds the cloud servers with the given serverIDs to the corresponding
+// load balancer pools with the given poolIDs.
+func CreateNodes(c *gophercloud.ServiceClient, opts []NodesOpts) CreateNodesResult {
+	var res CreateNodesResult
+	reqBody := make([]map[string]interface{}, len(opts))
+	for i := range opts {
```
jamiehannaford: maybe you could alias `NodesOpts` with `[]NodeOpts` and add a method that will do this serialisation for you? you'd also avoid added duplication

rackspace/rackconnect/v3/lbpools/requests.go
```diff
@@ -0,0 +1,156 @@
...
+	}
+	_, res.Err = c.Request("DELETE", createNodesURL(c), gophercloud.RequestOpts{
+		JSONBody: &reqBody,
+		OkCodes:  []int{204},
+	})
+	return res
+}
+
+// ListNodesDetailsForServer returns all load balancer pool nodes that are associated with RackConnect
+// for the given LB pool ID with all their details for the server with the given serverID.
```
jamiehannaford: I think this comment is a bit unclear

rackspace/rackconnect/v3/lbpools/requests_test.go
```diff
@@ -0,0 +1,876 @@
...
+	th.SetupHTTP()
+	defer th.TeardownHTTP()
+	th.Mux.HandleFunc("/load_balancer_pools/d6d3aa7c-dfa5-4e61-96ee-1d54ac1075d2/nodes", func(w http.ResponseWriter, r *http.Request) {
+		th.TestMethod(t, r, "GET")
+		th.TestHeader(t, r, "X-Auth-Token", fake.TokenID)
+		th.TestHeader(t, r, "Accept", "application/json")
+
+		w.Header().Set("Content-Type", "application/json")
+		fmt.Fprintf(w, `[
+      {
```
jamiehannaford: I don't mind, but is there a reason to not chuck all this data in a `fixtures.go`?

jamiehannaford: It'd be good if that helper fn which I added to my pending PR was merged then we could set up this kind of stuff easily :/

rackspace/rackconnect/v3/lbpools/results.go
```diff
@@ -0,0 +1,505 @@
...
+	// The cloud server (node) of the load balancer pool.
+	CloudServer struct {
+		// The cloud server ID.
+		ID string `mapstructure:"id"`
+	} `mapstructure:"cloud_server"`
+	// The load balancer pool.
+	LoadBalancerPool struct {
+		// The LB pool ID.
+		ID string `mapstructure:"id"`
+	} `mapstructure:"load_balancer_pool"`
```
jamiehannaford: is there a benefit to using anonymous structs like this?

rackspace/rackconnect/v3/cloudnetworks/requests.go
smashwilson: :+1: I'm on board with this.

Something else I've thought about is providing the ability to map other status codes to concrete errors, so a `Get` call could return `ErrContainerNotFound` instead of the more generic "unexpected status code" error. The benefit to doing this would be that it would be easier for client code to catch common error conditions and recover appropriately -- as another concrete example, it's currently awkward to report when authentication has failed.

Let's make a pair of issues about this :point_right:

smashwilson: Created as #400 and #401.

rackspace/rackconnect/v3/lbpools/requests.go
jrperritt: Named as such because it's specifically for operation on multiple nodes.

rackspace/rackconnect/v3/lbpools/requests_test.go
jrperritt: `fixtures.go` was mostly so that Rackspace packages could take advantage of OpenStack testing that was already setup. That doesn't really apply here.

rackspace/rackconnect/v3/lbpools/results.go
jrperritt: I don't think we need to have defined structures here. Users don't have to create anonymous structs like that, and each response is nuanced in what it returns, so we'd end up with structs that are the same but for one field.

rackspace/rackconnect/v3/lbpools/requests.go
jrperritt: Changed to be clearer.

rackspace/rackconnect/v3/lbpools/doc.go
jrperritt: :+1: Good catch. Copy and pasted from docs :/

rackspace/rackconnect/v3/cloudnetworks/requests.go
jrperritt: :+1:

rackspace/rackconnect/v3/lbpools/requests.go
jrperritt: On second thought, I'll change this.

rackspace/rackconnect/v3/lbpools/requests.go
jrperritt: :+1:

jrperritt:
Thank you for the review, @jamiehannaford :smile: 
:shipit:

Issue #400, opened 2015-03-18
Add default OkCodes to gophercloud.Request


smashwilson:
:bulb: from @jamiehannaford on [another PR](https://github.com/rackspace/gophercloud/pull/399#discussion_r26649123):

The status code that means "success" is mostly the same for all requests of each HTTP method. Let's make these the defaults and allow overriding when necessary.

PR #402, opened 2015-03-23
Add default HTTP statuses

jamiehannaford:
Addresses #400

### DIFF
--- a/provider_client.go
+++ b/provider_client.go
@@ -244,3 +244,57 @@ func defaultOkCodes(method string) []int {
 
 	return []int{}
 }
+
+func (client *ProviderClient) Get(url string, JSONResponse *interface{}, opts *RequestOpts) (*http.Response, error) {
+	if opts == nil {
+		opts = &RequestOpts{}
+	}
+	if JSONResponse != nil {
+		opts.JSONResponse = JSONResponse
+	}
+	return client.Request("GET", url, *opts)
+}
+
+func (client *ProviderClient) Post(url string, JSONBody interface{}, JSONResponse *interface{}, opts *RequestOpts) (*http.Response, error) {
+	if opts == nil {
+		opts = &RequestOpts{}
+	}
+
+	if v, ok := (JSONBody).(io.Reader); ok == true {
+		opts.RawBody = v
+	} else if JSONBody != nil {
+		opts.JSONBody = JSONBody
+	}
+
+	if JSONResponse != nil {
+		opts.JSONResponse = JSONResponse
+	}
+
+	return client.Request("POST", url, *opts)
+}
+
+func (client *ProviderClient) Put(url string, JSONBody interface{}, JSONResponse *interface{}, opts *RequestOpts) (*http.Response, error) {
+	if opts == nil {
+		opts = &RequestOpts{}
+	}
+
+	if v, ok := (JSONBody).(io.Reader); ok == true {
+		opts.RawBody = v
+	} else if JSONBody != nil {
+		opts.JSONBody = JSONBody
+	}
+
+	if JSONResponse != nil {
+		opts.JSONResponse = JSONResponse
+	}
+
+	return client.Request("PUT", url, *opts)
+}
+
+func (client *ProviderClient) Delete(url string, opts *RequestOpts) (*http.Response, error) {
+	if opts == nil {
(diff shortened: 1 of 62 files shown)


jamiehannaford:
@smashwilson @jrperritt Two things which I didn't introduce here, but I'd be interested to get everyone's opinions about:

1. Now we have a lot of empty `gophercloud.RequestOpts` structs. It'd be nice if they were optional, but we'd need to make them pointers if that's the case - which is a bit meh.
2. We've talked about it briefly before, but I think it'd be nice to add a few convenience functions to remove a lot of the verbosity for common http methods. I was thinking of:

```go
// URL and response struct required.
func (client *ProviderClient) Get(url string, JSONResponse interface{}, opts *RequestOpts) (*http.Response, error)
```
```go
// URL, JSON body and response required.
func (client *ProviderClient) Post(url string, JSONBody interface{}, JSONResponse interface{}, opts *RequestOpts) (*http.Response, error)
```
```go
// URL required.
func (client *ProviderClient) Delete(url string) (*http.Response, error)
```

smashwilson:
I'm on board with both of those changes.

Making `RequestOpts` a pointer means a lot (more) tiny changes to existing call sites to add that `&`, but I'd say it's worth it.

I believe `JSONResponse` would need to explicitly be a pointer (`*interface{}`) in `Get` and `Post`. Also, I think Perigee used to do exactly this, but deprecated the shortcut methods at some point. I'm not sure why, but it might be worth finding out before we do the same thing.

smashwilson:
> Also, I think Perigee used to do exactly this, but deprecated the shortcut methods at some point. I'm not sure why, but it might be worth finding out before we do the same thing.

Here we go: racker/perigee@a3f72a97

It looks like the general `Request` function used to be package-private, and *only* `Get`, `Post`, and so on were public. That commit deprecated the per-method entry points in favor of a single public function, with a consistent return value and a uniform way to report interpreted results. Reasonable.

Since we're *starting* with a public general method and adding the others as helpers, this should be fine for us :notes:

smashwilson:
@jamiehannaford This looks ready to merge as-is. Was there anything else you were going to do in this PR, or should I go ahead and :shipit:?

jamiehannaford:
@smashwilson Ok, pretty sure this is ready for review. 942 lines removed with ~300 lines added :metal: 

Since this PR pretty much touches everything, @jrperritt might also want to take a look

provider_client.go
```diff
@@ -226,3 +229,72 @@ func (client *ProviderClient) Request(method, url string, options RequestOpts) (
...
+		return []int{202, 204}
+	}
+
+	return []int{}
+}
+
+func (client *ProviderClient) Get(url string, JSONResponse *interface{}, opts *RequestOpts) (*http.Response, error) {
+	if opts == nil {
+		opts = &RequestOpts{}
+	}
```
smashwilson: Any reason to handle the `opts == nil` case in these utility methods instead of in `Request` itself? Since we're *already* touching most of the repository...

provider_client.go
```diff
@@ -226,3 +229,72 @@ func (client *ProviderClient) Request(method, url string, options RequestOpts) (
...
+	return client.Request("GET", url, *opts)
+}
+
+func (client *ProviderClient) Post(url string, JSONBody interface{}, JSONResponse *interface{}, opts *RequestOpts) (*http.Response, error) {
+	if opts == nil {
+		opts = &RequestOpts{}
+	}
+
+	if v, ok := (JSONBody).(io.Reader); ok == true {
```
smashwilson: `ok == true` is a bit redundant, you can just do `; ok {`. :grin:

smashwilson:
@jamiehannaford this is awesome, it really cleans a lot of things up :fireworks:

provider_client.go
jamiehannaford: hmm, initially I said no - but L252-254 sets the `JSONResponse` field, so it can't operate on `nil`

provider_client.go
jamiehannaford: :+1:

provider_client.go
smashwilson: Oh yeah, good point. Carry on, carry on.

smashwilson:
:shipit:

PR #403, opened 2015-03-26
[rfr] Adds Fixed IP support to os-floating-ips

jtopjian:
This commit enables the ability to specify a fixed IP when associating a
floating IP to an instance. If a fixed IP is not specified, Nova will
attempt to associate the floating IP to the first detected fixed IP, as it
did prior to this patch.

### DIFF
--- a/openstack/compute/v2/extensions/floatingip/requests.go
+++ b/openstack/compute/v2/extensions/floatingip/requests.go
@@ -100,7 +100,7 @@ func Delete(client *gophercloud.ServiceClient, id string) DeleteResult {
 // association / disassociation
 
 // Associate pairs an allocated floating IP with an instance
-// Deprecated. Use AssociateFloatingIP.
+// Deprecated. Use AssociateInstance.
 func Associate(client *gophercloud.ServiceClient, serverId, fip string) AssociateResult {
 	var res AssociateResult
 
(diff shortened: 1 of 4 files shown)


jtopjian:
Any update on this one? It changes the existing code quite a bit, so is there a better way to add this functionality without having to change the existing code so much?

jrperritt:
Sorry for the delay; I didn't know if this was ready for review. The only problem right now is backward compatibility. Maybe you could create new functions named `AssociateInstance` and `DisassociateInstance` (or something of the like) and then leave `Associate` and `Disassociate` as they are but add comments about them being deprecated in favor of the new functions. Does that make sense?

jtopjian:
Oops - my bad. I'll note if PRs are ready to go in the future. :)

Yes, that does make sense. I'll re-work this going that route. Thanks!

jrperritt:
:+1:

jtopjian:
@jrperritt Finally getting around to this!

Clean commit with a new approach. The existing functions, `Associate` and `Disassociate`, have been left alone. Two new functions, `AssociateFloatingIP` and `DisassociateFloatingIP`, have been created with the ability to specify a Fixed IP To pair the Floating IP with.

openstack/compute/v2/extensions/floatingip/requests.go
```diff
@@ -79,7 +112,33 @@ func Associate(client *gophercloud.ServiceClient, serverId, fip string) Associat
 	return res
 }
 
+// AssociateFloatingIP pairs an allocated floating IP with an instance.
+func AssociateFloatingIP(client *gophercloud.ServiceClient, opts AssociateOpts) AssociateResult {
```
jrperritt: Would it make sense to name this `AssociateInstance` so that when it called it would be `floatingip.AssociateInstance` instead of `floatingip.AssociateFloatingIP`? Similar proposal for `DisassociateFloatingIP`.

jtopjian: I agree. I debated this when I originally wrote it and decided to wait and see if someone else flagged it :)

jrperritt:
Thanks for sticking with this; I'm sorry it took me so long to circle back around to it. +2

jtopjian:
No problem at all! Thanks!

PR #406, opened 2015-03-31
ssh key name can be sent during server creation

einarf:
Looks like the `CreateOpts` for `openstack/compute/v2/servers` was missing a key pair name. Users should be able to easily supply the name of an exiting key during creation.

This is maybe not surprising since it's not documented :
https://bugs.launchpad.net/openstack-api-site/+bug/1300870

### DIFF
--- a/openstack/compute/v2/servers/requests.go
+++ b/openstack/compute/v2/servers/requests.go
@@ -131,6 +131,9 @@ type CreateOpts struct {
 
 	// ConfigDrive [optional] enables metadata injection through a configuration drive.
 	ConfigDrive bool
+
+	// KeyPair [optional] specifies an key pair for the instance for authentication 
+	KeyPairName string
 }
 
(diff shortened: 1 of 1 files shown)


ggiamarchi:
Actually keypairs are already supported. Gophercloud follows the OpenStack API design. Keypairs are not in the base compute API but in an [extension](http://developer.openstack.org/api-ref-compute-v2-ext.html). This is the same in [Gophercloud](https://github.com/rackspace/gophercloud/tree/master/openstack/compute/v2/extensions/keypairs).

You can see [here](https://github.com/docker/machine/blob/v0.2.0-rc2/drivers/openstack/client.go#L70) an example to see how to use it.

einarf:
Thanks. I don't know how I completely missed `CreateOptsExt`. Already using the extension to create key pairs. I guess we can close this one.

PR #408, opened 2015-03-31
Keystone v3 service catalog + auth scope

### LINKED ISSUE #407: Identity V3 and service catalog
Got an issue with how the service catalog is fetched in identity v3. The current implementation is looking up endpoints by listing services and endpoints through `/services` and `/endpoints`. Not all providers give users access to do a `identity:service_list` or `identity:endpoint_list`, so this will not work for them.

There might be very good reasons for the current implementation, but why not just use the catalog returned in the token and/or query `/auth/catalog` to obtain it? As long as you are scoped to a domain and/or project, the catalog should always be present unless you specifically supply and opt-out query.

https://github.com/openstack/keystone-specs/blob/master/api/v3/identity-api-v3.rst

I ended up making a small package to override stuff for now. I'll gladly make a PR, but always better to discuss here first.

ggiamarchi:
Fix #407 

Todo list :
 * [x] Read service catalog for keystone v3
 * [x] Build scope from auth opts for keystone v3 auth requests
 * [x] Fix unit tests

### DIFF
--- a/openstack/client.go
+++ b/openstack/client.go
@@ -133,10 +133,18 @@ func v3auth(client *gophercloud.ProviderClient, endpoint string, options gopherc
 		v3Client.Endpoint = endpoint
 	}
 
-	token, err := tokens3.Create(v3Client, options, nil).Extract()
+	result := tokens3.Create(v3Client, options, nil)
+
+	token, err := result.ExtractToken()
 	if err != nil {
 		return err
 	}
+
+	catalog, err := result.ExtractServiceCatalog()
+	if err != nil {
+		return err
+	}
+
 	client.TokenID = token.ID
 
 	if options.AllowReauth {
@@ -145,7 +153,7 @@ func v3auth(client *gophercloud.ProviderClient, endpoint string, options gopherc
 		}
 	}
 	client.EndpointLocator = func(opts gophercloud.EndpointOpts) (string, error) {
-		return V3EndpointURL(v3Client, opts)
+		return V3EndpointURL(catalog, opts)
 	}
 
 	return nil
--- a/openstack/client.go
+++ b/openstack/client.go
@@ -133,7 +133,25 @@ func v3auth(client *gophercloud.ProviderClient, endpoint string, options gopherc
 		v3Client.Endpoint = endpoint
 	}
 
-	result := tokens3.Create(v3Client, options, nil)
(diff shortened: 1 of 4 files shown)


einarf:
Nice. Pretty much identical to what I did locally with the exception of the `v3auth` changes. +1 for that. There was never a need to generate two tokens on every auth.

jrperritt:
@ggiamarchi Thank you for the PR. Would it be possible to implement this while remaining backwards-compatible? For instance:
https://github.com/ggiamarchi/gophercloud/blob/keystone-v3-auth/openstack/endpoint_location.go#L59
vs
https://github.com/rackspace/gophercloud/blob/master/openstack/endpoint_location.go#L61

ggiamarchi:
@jrperritt Unfortunately, i don't see how to be backwards-compatible in this specific case. To keep the method signature unchanged we should make API calls to read the service catalog from inside the function `V3EndpointURL`. In the v3 API, the only reliable way to get the service catalog for a standard user is to read it from the authentication response. So, the only solution i found is to pass the service catalog as a parameter.

Furthermore, IMO this is not very annoying in this case to break the backward compatibility because the current implementation actually does not work and this method is more intended to be called internally in gophercloud.

However, if you see a solution to make it work with a backward-compatible approach, let me know.

einarf:
@ggiamarchi is correct. I don't think there is any point supporting the old method of extracting the catalog unless someone for some odd reason have a custom setup that actually returns anything that looks like the service catalog from `/services` and `/endpoints'. The services and endpoints packages that were added in the original commit are still useful though.

Endpoints returns a structure like this :
```
[
    {
        "id": "--endpoint-id--",
        "interface": "public",
        "links": {
            "self": "http://identity:35357/v3/endpoints/--endpoint-id--"
        },
        "name": "the public volume endpoint",
        "service_id": "--service-id--"
    },
    {
        "id": "--endpoint-id--",
        "interface": "internal",
        "links": {
            "self": "http://identity:35357/v3/endpoints/--endpoint-id--"
        },
        "name": "the internal volume endpoint",
        "service_id": "--service-id--"
    }
]
```

jrperritt:
OK, makes sense :+1:

einarf:
Working like a dream!

ggiamarchi:
:ok_hand: :smiley:

PR #410, opened 2015-04-07
Remove leading "v" from user-agent string

jamiehannaford:
FAO @everett-toews @jrperritt @smashwilson


smashwilson:
:+1:

everett-toews:
Thanks!

PR #412, opened 2015-04-08
More Rackspace servers updates

jrperritt:
This PR has 2 changes:
1. `os.ResizeOpts` -> `os.ResizeOptsBuilder` for resizing a Rackspace server
2. Add `AccessIPv4` and `AccessIPv6` fields as options when creating a Rackspace server


rackspace/compute/v2/servers/requests.go
```diff
@@ -47,6 +47,12 @@ type CreateOpts struct {
 	// password will be created and returned in the response.
 	AdminPass string
 
+	// AccessIPv4 [optional] specifies an IPv4 address for the instance.
+	AccessIPv4 string
```
smashwilson: Wait, you can do that? Huh.

jrperritt: Seems not. Oh well.

jrperritt:
:shipit:

PR #413, opened 2015-04-10
Server personality

KevinPike:
This pull request changes server Personality from `[]byte` to `[]map[string]string`, where each map is expected to contain path and base64 encoded contents.


mdarby:
@jrperritt could you please review?

openstack/compute/v2/servers/requests.go
```diff
@@ -125,9 +125,10 @@ type CreateOpts struct {
 	// Metadata [optional] contains key-value pairs (up to 255 bytes each) to attach to the server.
 	Metadata map[string]string
 
-	// Personality [optional] includes the path and contents of a file to inject into the server at launch.
-	// The maximum size of the file is 255 bytes (decoded).
-	Personality []byte
+	// Personality [optional] includes a list of maps with the path and contents
+	// of a file to inject into the server at launch. Contents should be
+	// base64 encoded. The maximum size of the file is 255 bytes (decoded).
+	Personality []map[string]string
```
jrperritt: Since it's fixed to be only `path` and `contents`, maybe we can have a `Personality` struct:
```go
type Personality struct {
    Path string
    Contents string
}
```

and then have the field definition in the `CreateOpts` (and the like) be
``` Personality []Personality```

**Comment by jrperritt** [comment]
@KevinPike Great catch. Just one suggestion above.

File: openstack/compute/v2/servers/requests.go (hunk 1 above)
Comment (KevinPike) [inline openstack/compute/v2/servers/requests.go]: Good call. Perhaps `Contents` could be `[]byte` and base64 encoded on Create and Rebuild, similar to how `UserData` is base64 encoded on Create.
```go
type Personality struct {
    Path string
    Contents []byte
}
```

jrperritt: :+1:

KevinPike:
I've tested this with Ubuntu and it works. However, I haven't had luck with Windows.

This [post](https://developer.rackspace.com/blog/step-by-step-walkthrough-to-using-chef-to-bootstrap-windows-nodes-on-the-rackspace-cloud/) uses personality with a file path `"C:\\cloud-automation\\bootstrap.cmd"` to create a Windows server and it works. 

I don't think it's working on Windows because in Go `"C:\\cloud-automation\\bootstrap.cmd"` serializes to `"C:\\cloud-automation\\bootstrap.cmd"`. In Python, it serializes to `"C:\cloud-automation\bootstrap.cmd"`.  Any thoughts?

jrperritt:
I don't know much (read: anything) about Windows environments, but maybe you can use a forward-slash instead? For example, "C:/cloud-automation/bootstrap.cmd".

KevinPike:
It turns out that `"C:\\cloud-automation\\bootstrap.cmd"` is a valid path and the file will be created when using personality. I was using `"C:\\cloud-automation\\bootstrap.ps1"` which did not work, interestingly. 

I updated Personality to marshal using MarshalJSON.

I've tested this on Ubuntu 14.04 and Windows Server 2012.

KevinPike:
Ready for review

openstack/compute/v2/servers/requests.go
```diff
@@ -95,6 +96,29 @@ type Network struct {
...
+// File is used within CreateOpts and RebuildOpts to inject a file into the server at launch.
+type File struct {
+	// Path of the file
+	Path string
+	// Contents of the file. Maximum content size is 255 bytes.
+	Contents []byte
+}
+
+// MarshalJSON marshals the escaped file, base64 encoding the contents.
+func (f *File) MarshalJSON() ([]byte, error) {
```
jrperritt: What is the intended use case of the method? It seems like the base64 encoding process should go in the `ToServerCreateMap` method.

KevinPike: The file base64 encode process is used by `ToServerCreateMap` and `ToServerRebuildMap`. Having `File` implement `MarshalJSON` seemed like a good way to reduce putting the encode logic in two places and have a `File` know how to marshal itself.

jrperritt: Oh, I see; it's implementing the `Marshaler` interface. Can you add a comment above `File` mentioning that? Something like: "File implements the `json.Marshaler` interface, so when a `Create` or `Rebuild` operation is requested, `json.Marshal` will call `File`'s `MarshalJSON method." The only reason for explicitness is because I don't think we have this anywhere else. Also, could you add a `Personality` field to the acceptance test that creates a server?

KevinPike: Sure! :+1:

jrperritt:
It looks like the unit tests are failing. From Travis:
`FAIL	github.com/rackspace/gophercloud/openstack/compute/v2/servers [build failed]`

KevinPike:
There we go. Thank you for the heads up

jrperritt:
Looks good. Thank you for the PR :smile: 

 :rocket:

PR #414, opened 2015-04-11
Updating paths from code.google.com/p/* to golang.org/x/*

jackspirou:
Hi, 

Quick comment.

As I was dealing with some dependency management in `gophercloud` earlier this week, I noticed that one of the test files is still utilizing `code.google.com/p/go.crypto` instead of `golang.org/x/crypto`.  

As everyone is this project is probably aware, the golang team has moved these dependencies from `code.google.com/p/*` to `golang.org/x/*`.  You can find the google groups announcement here: https://groups.google.com/forum/#!topic/golang-nuts/eD8dh3T9yyA

If we can switch over to `golang.org/x/` I would propose we do so.  It would help people like me who are dealing with their own golang dependency management and are now expecting the new import paths.

Thanks!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ go:
   - tip
 script: script/cibuild
 after_success:
-  - go get code.google.com/p/go.tools/cmd/cover
+  - go get golang.org/x/tools/cmd/cover
   - go get github.com/axw/gocov/gocov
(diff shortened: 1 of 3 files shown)


jrperritt:
Ugh, I've been so focused on adding features that I'd lost sight of maintenance. Thank you for fixing this @jackspirou +1

PR #416, opened 2015-04-13
feature - added ability to retreive any attached volume from an instance

clintonskitson:
When doing a Get request on attached volumes of an instance, it was previously required that you enter a volume ID.  These small changes allow you to not specify a volume ID, and return a proper list of currently attached volumes.

### DIFF
--- a/openstack/compute/v2/extensions/volumeattach/results.go
+++ b/openstack/compute/v2/extensions/volumeattach/results.go
@@ -65,6 +65,22 @@ func (r VolumeAttachmentResult) Extract() (*VolumeAttachment, error) {
 	return res.VolumeAttachment, err
 }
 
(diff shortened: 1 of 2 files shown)


jrperritt:
I'm sure I'm missing something obvious, but is this not just the [`List`](https://github.com/rackspace/gophercloud/blob/master/openstack/compute/v2/extensions/volumeattach/requests.go#L11) function? For example, the following would give you all the volumes attached to an instance (without pagination, which is also possible):
``` go
allPages, err := volumeattach.List(client, serverID).AllPages()
if err != nil {
    // handle error...
}
allAttachedVols, err := volumeattach.ExtractVolumeAttachments(allPages)
```

clintonskitson:
Yup, you're right.  Thanks for the quick reply.

PR #417, opened 2015-04-14
add image.Delete operation

spheromak:
This operation is documented in the docs, but was not implemented. So lets implement it!

http://gophercloud.io/docs/compute/#delete-image

### DIFF
--- a/openstack/compute/v2/images/requests.go
+++ b/openstack/compute/v2/images/requests.go
@@ -63,3 +63,10 @@ func Get(client *gophercloud.ServiceClient, id string) GetResult {
 	_, result.Err = client.Get(getURL(client, id), &result.Body, nil)
 	return result
 }
+
+// Delete deletes the specified image ID.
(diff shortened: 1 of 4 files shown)


jrperritt:
Ah, quite right, thank you for the PR :)  Just a few suggestions to make it a bit more consistent with the other functions:

2. Adding a [unit test](https://github.com/rackspace/gophercloud/blob/master/openstack/compute/v2/images/requests_test.go)
3. Adding a function called `deleteURL` (in [`urls.go`](https://github.com/rackspace/gophercloud/blob/master/openstack/compute/v2/images/urls.go)) and using it instead of `getURL`. No need to add a unit test for `deleteURL`; We've deprecated that practice :)

spheromak:
I went to add a test, but I don't understand what test I could add here to add value. Everything that is being used is already covered via the middleware functions. we simply take the string throw it against the api and return any error.

spheromak:
@jrperritt fixed up per your feedback

jrperritt:
Thank you :smile: 

:rocket:

PR #419, opened 2015-04-16
Server actions : CreateImage

einarf:
A server action for creation a sever image was missing. This is identical to how the standard openstack client operates when doing a "server image create". Might be a bit quick and dirty right now, but works.

Need something like this to get proper gophercloud support in the Packer project.


einarf:
Forgot to update some tests. Fixing that tomorrow.

einarf:
Closing this PR and making a new one in a bit. This one is way too messy.

PR #421, opened 2015-04-16
server action to create a server image / snapshot

einarf:
Missing this server action to create a snapshot/image from a server. This is identical to what the openstack client does after a `server image create`.

### DIFF
--- a/openstack/compute/v2/servers/requests.go
+++ b/openstack/compute/v2/servers/requests.go
@@ -4,7 +4,6 @@ import (
 	"encoding/base64"
 	"errors"
 	"fmt"
-	"strings"
 
 	"github.com/rackspace/gophercloud"
 	"github.com/rackspace/gophercloud/pagination"
@@ -702,23 +701,23 @@ func ListAddressesByNetwork(client *gophercloud.ServiceClient, id, network strin
 	return pagination.NewPager(client, listAddressesByNetworkURL(client, id, network), createPageFn)
 }
 
-type CreateServerImageOpts struct {
+type CreateImageOpts struct {
 	// Name [required] of the image/snapshot
 	Name string
 	// Metadata [optional] contains key-value pairs (up to 255 bytes each) to attach to the created image.
 	Metadata map[string]string
 }
 
-type CreateServerImageOptsBuilder interface {
-	ToCreateServerImageMap() (map[string]interface{}, error)
+type CreateImageOptsBuilder interface {
+	ToServerCreateImageMap() (map[string]interface{}, error)
 }
 
-// ToServerImageCreateMap formats an ServerImageCreateOpts structure into a request body.
-func (opts CreateServerImageOpts) ToCreateServerImageMap() (map[string]interface{}, error) {
+// ToServerCreateImageMap formats a CreateImageOpts structure into a request body.
+func (opts CreateImageOpts) ToServerCreateImageMap() (map[string]interface{}, error) {
 	var err error
 	img := make(map[string]interface{})
 	if opts.Name == "" {
-		err = fmt.Errorf("Cannot create a server image without a name")
+		return nil, fmt.Errorf("Cannot create a server image without a name")
 	}
 	img["name"] = opts.Name
 	if opts.Metadata != nil {
@@ -729,30 +728,10 @@ func (opts CreateServerImageOpts) ToCreateServerImageMap() (map[string]interface
 	return createImage, err
 }
 
-// ExtractImageID gets the ID of the newly created server image from the header
-func (res CreateServerImageResult) ExtractImageID() (string, error) {
-	var err error
-	if res.Err != nil {
-		return "", res.Err
-	}
-	// Get the image id from the header
-	rawUrl := res.Header.Get("Location")
-	fmt.Println("RawUrl:", rawUrl)
-	if rawUrl != "" {
-		split := strings.Split(rawUrl, "/")
-		if len(split) > 0 {
-			id := split[len(split)-1]
-			return id, nil
-		}
-	}
-	err = fmt.Errorf("Failed to parse the ID of newly created image")
-	return "", err
-}
-
-// CreateServerImage makes a request against the nova API to schedule an image to be created of the server
-func CreateServerImage(client *gophercloud.ServiceClient, serverId string, opts CreateServerImageOptsBuilder) CreateServerImageResult {
-	var res CreateServerImageResult
-	reqBody, err := opts.ToCreateServerImageMap()
+// CreateImage makes a request against the nova API to schedule an image to be created of the server
+func CreateImage(client *gophercloud.ServiceClient, serverId string, opts CreateImageOptsBuilder) CreateImageResult {
+	var res CreateImageResult
+	reqBody, err := opts.ToServerCreateImageMap()
 	if err != nil {
 		res.Err = err
 		return res
@@ -761,8 +740,6 @@ func CreateServerImage(client *gophercloud.ServiceClient, serverId string, opts 
 		OkCodes: []int{202},
 	})
 	res.Err = err
-	if err == nil {
-		res.Header = response.Header
-	}
+	res.Header = response.Header
 	return res	
 }
--- a/openstack/compute/v2/servers/requests.go
+++ b/openstack/compute/v2/servers/requests.go
@@ -4,6 +4,7 @@ import (
 	"encoding/base64"
 	"errors"
 	"fmt"
+	"strings"
 
 	"github.com/rackspace/gophercloud"
 	"github.com/rackspace/gophercloud/pagination"
@@ -14,7 +15,6 @@ import (
 type ListOptsBuilder interface {
 	ToServerListQuery() (string, error)
 }
-
 // ListOpts allows the filtering and sorting of paginated collections through
 // the API. Filtering is achieved by passing in struct field values that map to
 // the server attributes you want to see returned. Marker and Limit are used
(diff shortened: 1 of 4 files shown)


openstack/compute/v2/servers/requests.go
```diff
@@ -701,3 +701,68 @@ func ListAddressesByNetwork(client *gophercloud.ServiceClient, id, network strin
 	}
 	return pagination.NewPager(client, listAddressesByNetworkURL(client, id, network), createPageFn)
 }
+
+type CreateServerImageOpts struct {
```
jrperritt: s/CreateServerImageOpts/CreateImageOpts

Since we're in the `servers` package, we can drop the 'Server' part of the names. It'll also be more consistent with the rest of the package.

openstack/compute/v2/servers/requests.go
```diff
@@ -701,3 +701,68 @@ func ListAddressesByNetwork(client *gophercloud.ServiceClient, id, network strin
...
+	// Name [required] of the image/snapshot
+	Name string
+	// Metadata [optional] contains key-value pairs (up to 255 bytes each) to attach to the created image.
+	Metadata map[string]string
+}
+
+type CreateServerImageOptsBuilder interface {
```
jrperritt: s/CreateServerImageOptsBuilder/CreateImageOptsBuilder

openstack/compute/v2/servers/requests.go
```diff
@@ -701,3 +701,68 @@ func ListAddressesByNetwork(client *gophercloud.ServiceClient, id, network strin
...
+	ToCreateServerImageMap() (map[string]interface{}, error)
```
jrperritt: s/ToCreateServerImageMap/ToServerCreateImageMap

openstack/compute/v2/servers/requests.go
```diff
@@ -701,3 +701,68 @@ func ListAddressesByNetwork(client *gophercloud.ServiceClient, id, network strin
...
+}
+
+// ToServerImageCreateMap formats an ServerImageCreateOpts structure into a request body.
```
jrperritt: s/ToServerImageCreateMap/ToServerCreateImageMap
s/an/a
s/ServerImageCreateOpts/CreateImageOpts

openstack/compute/v2/servers/requests.go
```diff
@@ -701,3 +701,68 @@ func ListAddressesByNetwork(client *gophercloud.ServiceClient, id, network strin
...
+func (opts CreateServerImageOpts) ToCreateServerImageMap() (map[string]interface{}, error) {
```
jrperritt: s/CreateServerImageOpts/CreateImageOpts
s/ToCreateServerImageMap/ToServerCreateImageMap

openstack/compute/v2/servers/requests.go
```diff
@@ -701,3 +701,68 @@ func ListAddressesByNetwork(client *gophercloud.ServiceClient, id, network strin
...
+	var err error
+	img := make(map[string]interface{})
+	if opts.Name == "" {
+		err = fmt.Errorf("Cannot create a server image without a name")
```
jrperritt: Good check for a required field. Let's return early if this isn't set:
```go
if opts.Name == "" {
    return nil, fmt.Errorf("Cannot create a server image without a name")
}
```

openstack/compute/v2/servers/requests.go
```diff
@@ -701,3 +701,68 @@ func ListAddressesByNetwork(client *gophercloud.ServiceClient, id, network strin
...
+	if opts.Metadata != nil {
+		img["metadata"] = opts.Metadata
+	}
+	createImage := make(map[string]interface{})
+	createImage["createImage"] = img
+	return createImage, err
+}
+
+// ExtractImageID gets the ID of the newly created server image from the header
+func (res CreateServerImageResult) ExtractImageID() (string, error) {
```
jrperritt: Let's move this function into `results.go` directly beneath the `CreateImageResult` definition.

openstack/compute/v2/servers/requests.go
```diff
@@ -701,3 +701,68 @@ func ListAddressesByNetwork(client *gophercloud.ServiceClient, id, network strin
...
+	var err error
+	if res.Err != nil {
+		return "", res.Err
+	}
+	// Get the image id from the header
+	rawUrl := res.Header.Get("Location")
+	fmt.Println("RawUrl:", rawUrl)
+	if rawUrl != "" {
+		split := strings.Split(rawUrl, "/")
+		if len(split) > 0 {
```
jrperritt: We might want to do a different check here. I think this will always evaluate to `true`.

einarf:
Looking into it right now.

openstack/compute/v2/servers/requests.go
```diff
@@ -701,3 +701,68 @@ func ListAddressesByNetwork(client *gophercloud.ServiceClient, id, network strin
...
+			id := split[len(split)-1]
+			return id, nil
+		}
+	}
+	err = fmt.Errorf("Failed to parse the ID of newly created image")
+	return "", err
```
jrperritt: We can git rid of `err` if we're only using it here and just:
 ```go
return "", fmt.Errorf("Failed to parse the ID of newly created image")
```

File: openstack/compute/v2/servers/requests.go [hunk 10]
```diff
@@ -701,3 +701,68 @@ func ListAddressesByNetwork(client *gophercloud.ServiceClient, id, network strin
...
+}
+
+// CreateServerImage makes a request against the nova API to schedule an image to be created of the server
```
jrperritt: s/CreateServerImage/CreateImage
s/to be created of the server/of the server to be created.

openstack/compute/v2/servers/requests.go
```diff
@@ -701,3 +701,68 @@ func ListAddressesByNetwork(client *gophercloud.ServiceClient, id, network strin
...
+func CreateServerImage(client *gophercloud.ServiceClient, serverId string, opts CreateServerImageOptsBuilder) CreateServerImageResult {
```
jrperritt: s/CreateServerImage/CreateImage
s/opts CreateServerImageOptsBuilder/opts CreateImageOptsBuilder
s/CreateServerImageResult/CreateImageResult

openstack/compute/v2/servers/requests.go
```diff
@@ -701,3 +701,68 @@ func ListAddressesByNetwork(client *gophercloud.ServiceClient, id, network strin
...
+	var res CreateServerImageResult
```
jrperritt: s/CreateServerImageResult/CreateImageResult

openstack/compute/v2/servers/requests.go
```diff
@@ -701,3 +701,68 @@ func ListAddressesByNetwork(client *gophercloud.ServiceClient, id, network strin
...
+	reqBody, err := opts.ToCreateServerImageMap()
```
jrperritt: s/ToCreateServerImageMap/ToServerCreateImageMap

openstack/compute/v2/servers/requests.go
```diff
@@ -701,3 +701,68 @@ func ListAddressesByNetwork(client *gophercloud.ServiceClient, id, network strin
...
+	if err != nil {
+		res.Err = err
+		return res
+	}
+	response, err := client.Post(actionURL(client, serverId), reqBody, nil, &gophercloud.RequestOpts{
+		OkCodes: []int{202},
+	})
+	res.Err = err
+	if err == nil {
```
jrperritt: No need to check for a `nil` error here; The client code will do that. Let's populate the header regardless of the value of `err`.

openstack/compute/v2/servers/requests_test.go
```diff
@@ -325,3 +325,13 @@ func TestListAddressesByNetwork(t *testing.T) {
 	th.AssertNoErr(t, err)
 	th.CheckEquals(t, 1, pages)
 }
+
+func TestCreateServerImage(t *testing.T) {
+	th.SetupHTTP()
+	defer th.TeardownHTTP()
+	HandleCreateServerImageSuccessfully(t)
+
+	_, err := CreateServerImage(client.ServiceClient(), "serverimage", CreateServerImageOpts{Name: "test"}).ExtractImageID()
```
jrperritt: s/CreateServerImage/CreateImage
s/CreateServerImageOpts/CreateImageOpts

openstack/compute/v2/servers/results.go
```diff
@@ -74,6 +74,11 @@ type RescueResult struct {
 	ActionResult
 }
 
+// RescueResult represents the result of a server rescue operation
```
jrperritt: s/RescueResult/CreateImageResult
s/server rescue/create image

openstack/compute/v2/servers/results.go
```diff
@@ -74,6 +74,11 @@ type RescueResult struct {
...
+type CreateServerImageResult struct {
```
jrperritt: s/CreateServerImageResult/CreateImageResult

jrperritt:
Thank you for this PR; We do need this operation. Most of my comments were about Gophercloud's naming conventions. And then there was one about splitting the `Location` header to get the image ID. They should be quick fixes.

einarf:
Great. Commit just around the corner. Having this server action is the last thing needed for the Packer project to use this API. Pretty much ready with a PR for that now.

jrperritt:
:+1:  :rocket:

einarf:
Awesome. That was the last missing piece in getting the new Gophercloud API to work for the Packer project. Time to do a huge PR there. Exited! :smiley:

PR #422, opened 2015-04-17
Scheduler Hints

jtopjian:
This commit adds the ability to specify scheduling hints for the
instances. Scheduling hints alter the placement of instances in
the OpenStack cloud by using a variety of attributes.

### DIFF
--- a/openstack/compute/v2/extensions/schedulerhints/requests.go
+++ b/openstack/compute/v2/extensions/schedulerhints/requests.go
@@ -1,0 +1,134 @@
+package schedulerhints
+
+import (
+	"fmt"
+	"net"
+	"regexp"
+	"strings"
+
+	"github.com/rackspace/gophercloud/openstack/compute/v2/servers"
+)
+
+// SchedulerHints represents a set of scheduling hints that are passed to the
+// OpenStack scheduler
+type SchedulerHints struct {
+	// Group specifies a Server Group to place the instance in.
+	Group string
+
+	// DifferentHost will place the instance on a compute node that does not
+	// host the given instances.
+	DifferentHost []string
+
+	// SameHost will place the instance on a compute node that hosts the given
+	// instances.
+	SameHost []string
+
+	// Query is a conditional statement that results in compute nodes able to
+	// host the instance.
+	Query []string
+
+	// TargetCell specifies a cell name where the instance will be placed.
+	TargetCell string
+
+	// BuildNearHostIP specifies a subnet of compute nodes to host the instance.
+	BuildNearHostIP string
+}
+
+// SchedulerHintsBuilder builds the scheduler hints into a serializable format.
+type SchedulerHintsBuilder interface {
+	ToServerSchedulerHintsMap() (map[string]interface{}, error)
+}
+
+// ToServerSchedulerHintsMap builds the scheduler hints into a serializable format.
+func (opts SchedulerHints) ToServerSchedulerHintsMap() (map[string]interface{}, error) {
+	sh := make(map[string]interface{})
+
+	uuidRegex, _ := regexp.Compile("^[a-z0-9]{8}-[a-z0-9]{4}-[1-5][a-z0-9]{3}-[a-z0-9]{4}-[a-z0-9]{12}$")
+
+	if opts.Group != "" {
+		if !uuidRegex.MatchString(opts.Group) {
+			return nil, fmt.Errorf("Group must be a UUID")
+		}
+		sh["group"] = opts.Group
+	}
+
+	if len(opts.DifferentHost) > 0 {
+		for _, diffHost := range opts.DifferentHost {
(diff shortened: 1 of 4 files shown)


jtopjian:
Ready for review.

jrperritt:
Great. I'll review tomorrow.

jrperritt:
Looking good. I have 2 observations:

1. I think this is an OpenStack extension, and, if so, we'll want to put it under the `extensions` directory.
2. It looks like the available fields to provide as scheduler hints may be fixed. See [here](https://github.com/openstack/nova/blob/master/nova/api/openstack/compute/schemas/v3/scheduler_hints.py). I acknowledge that link is for the new version of Nova, but if you look [here](http://docs.openstack.org/juno/config-reference/content/section_compute-scheduler.html), the only examples provided uses the values from that fixed schema. What do you think?

jtopjian:
1. Ugh. I must have had the extensions doc opened thinking it was the standard api doc. I'll recreate this as an extension similar to keypair. I remember being surprised that it wasn't an extension :)
2. Yes, I see your point. I'll admit that I'm siding on laziness here, but a few counterpoints:

* Wouldn't allowing arbitrary hints enable automatic future hints to work as well as potential removed hints to still work on older releases?
* Is this kind of the same as the security group `protocol` field?

If it's ultimately better to model the available hints, I'm open to doing the work, though.

jrperritt:
- Sure, having an general `map[string]interface{}` would allow for the most flexibility, but if there is a fixed schema, then we'll want to implement it instead. In addition, if you provide users with a `map[string]interface`, how are they to know what the valid options are?
- I'm not sure what you mean. Do you have a link?

How about this: We can use the defined schema as the `SchedulerHintOpts` but make the field in the `CreateOptsExt` of type `SchedulerHintOptsBuilder` (an `interface`). The only method that needs to be implemented to satisfy it is `ToServerSchedulerHintMap`, and it returns a `map[string]interface{}`. That way, users can use our defined `SchedulerHintOpts` `struct` or they can provide their own type that implements `SchedulerHintOptsBuilder`. Does that make sense?

jtopjian:
* Yeah, that's the tradeoff: I think invalid hints are silently ignored, so implementing some form of verification in the tool is helpful, but ... work :)
* Oh, I meant how there's not really a protocol enforcement for tcp, udp, icmp, etc. That's left for the user to choose. Unless I'm mistaken?

It does make sense. I'll give it a shot and see how far I get. Thank you for the input :)

jtopjian:
Here's a quick attempt at creating an extension. It's a bit different than the others I made since it extends the ServerCreate opts, doesn't have its own HTTP URL, and doesn't return anything... I copied heavily from the `diskconfig` extension.

Does this look correct? If so, I'll move on to modelling the Schedule Hints schema.

jtopjian:
Cool - thanks for the comments. I worked on the generic `map[string]interface{}`-to-struct conversion a few days ago and I believe satisfies the comments.

Short of that, does the structure of the extension look OK? I wanted to get that sorted out first.

jrperritt:
Yeah, looks fine to me.

jtopjian:
OK, cool. Let me know how this looks when you have time :smile:

openstack/compute/v2/extensions/schedulerhints/requests.go
```diff
@@ -0,0 +1,134 @@
...
+
+	return sh, nil
+}
+
+// CreateOptsExt adds a SchedulerHints option to the base CreateOpts.
+type CreateOptsExt struct {
+	servers.CreateOptsBuilder
+
+	// SchedulerHints provides a set of hints to the scheduler.
+	SchedulerHintsBuilder
```
jrperritt: Let's have an explicit field name here. How about `SchedulerHints`?

jrperritt:
Looks really good. Just one change to make (for consistency with the other extensions).

jtopjian:
Thanks for pointing that out. How does it look now?

openstack/compute/v2/extensions/schedulerhints/requests.go
```diff
@@ -0,0 +1,134 @@
...
+
+	return sh, nil
+}
+
+// CreateOptsExt adds a SchedulerHints option to the base CreateOpts.
+type CreateOptsExt struct {
+	servers.CreateOptsBuilder
+
+	// SchedulerHints provides a set of hints to the scheduler.
+	SchedulerHints SchedulerHintsBuilder
```
jrperritt: :+1:

jrperritt:
I have just one more small (and selfish) request: After a PR review has begun, can you append commits instead of rebasing? It'll be easier for the reviewer to see only what's changed since the last review. :smile: 

Regardless of that, this looks great. Thank you all the hard work :rocket:

jtopjian:
Sure - understood.

Thank you!

PR #425, opened 2015-04-27
Use a ReadSeeker for upload content and rewind the offset on retries

b0d0nne11:
I have an issue where if a re-authentication is required during a object upload that object is eventually created with no content. Here are the Cloud Files access logs from one failed upload. The initial response is a 401 unauthorized. The client then re-authenticates and successfully retries the request but silently fails to include the file content.

```
209.61.148.41 - - [27/04/2015:11:06:51 +0000] "PUT /v1/[redacted]/events-1430043166780366665.gz HTTP/1.0" 401 131 "-" "gophercloud/v1.0"
209.61.148.41 - - [27/04/2015:11:06:52 +0000] "PUT /v1/[redacted]/events-1430043166780366665.gz HTTP/1.0" 201 0 "-" "gophercloud/v1.0"
```

I believe the second time the request is called the `io.Reader` `options.RawBody` has no un-read content. Unfortunately the offset in `io.Reader` cant be rewound AFAIK. This PR resolves the issue by changing `options.RawBody` to a `io.ReadSeeker` and resetting the offset before re-trying requests.

### DIFF
--- a/openstack/objectstorage/v1/objects/requests.go
+++ b/openstack/objectstorage/v1/objects/requests.go
@@ -188,7 +188,7 @@ func (opts CreateOpts) ToObjectCreateParams() (map[string]string, string, error)
 }
 
 // Create is a function that creates a new object or replaces an existing object.
-func Create(c *gophercloud.ServiceClient, containerName, objectName string, content io.Reader, opts CreateOptsBuilder) CreateResult {
+func Create(c *gophercloud.ServiceClient, containerName, objectName string, content io.ReadSeeker, opts CreateOptsBuilder) CreateResult {
 	var res CreateResult
 
 	url := createURL(c, containerName, objectName)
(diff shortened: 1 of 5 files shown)


b0d0nne11:
Ready for review.

jrperritt:
Hmm, I think you're right about that. The `io.Reader` was in place before we implemented automatic re-auth. This wouldn't be completely backwards-compatible (for example, with `bytes.Buffer`), but it would be for what I'd guess is the overwhelming majority of cases: `os.File`, `strings`, and `bytes`. This change seems reasonable to me. @jamiehannaford @smashwilson Any thoughts on this?

b0d0nne11:
I double checked the documentation before I submitted this and made sure all the examples at http://gophercloud.io/docs/object-storage/#upload-object would still work. That page would still have to be updated to specifically call out the updated interface. I'm not to sure how to do that.

jrperritt:
Yeah, that site is created from the `gh-pages` branch: https://github.com/rackspace/gophercloud/blob/gh-pages/docs/object-storage/index.md#upload-object. I can update that. Thank you for the PR :smile: 

:rocket:

PR #426, opened 2015-05-05
Allows for more complex scheduling queries.

jtopjian:


jtopjian:
I noticed that I restricted query to `[]string` in the original PR and that nixed any way of creating a more complex query.

jrperritt:
:+1:

jtopjian:
Thanks!

PR #428, opened 2015-05-18
fix object storage nil pointer dereference error; Closes #427

### LINKED ISSUE #427: nil pointer dereference
https://github.com/rackspace/gophercloud/blob/master/openstack/objectstorage/v1/objects/requests.go#L138

In this line resp.Body is read. The gophercloud.ServiceClient returns "nil, err" in the error case, aka this will fail miserably. Better check for the error, like:
~~~
res.err = err
if resp != nil {
  res.Body = resp.Body
  res.Header = resp.Header
}
~~~

jrperritt:
If a request errors out, then the `resp` returned from the `gophercloud.Request` method will be `nil`. In that case, trying to access the `Header` field will fail. This PR tries to fix that.


jrperritt:
Same 1 change for 5 different files -> :shipit:

PR #429, opened 2015-05-20
create server helper fields

jrperritt:
This PR lets users create servers by providing image and flavor names (instead of having to provide IDs).


jrperritt:
hmm, it's odd that the formatting was only messed up for those 2 files. Oh, well. Thank you, @jamiehannaford  :smile: 

:shipit:

PR #430, opened 2015-06-04
IDFromName for servers

jrperritt:


jrperritt:
This is just like the `flavors` and `images` functions of the same name :shipit:

PR #437, opened 2015-06-24
Rackspace CLB ExtractNodes call does not propagate errors correctly

smashwilson:
I hit this playing around with rgbkrk/peekaboo#6. If a `nodes.Create` call fails, the result type's `ExtractNodes` method does not correctly return the error:

```
2015/06/24 13:23:19 Setting 10.209.224.145:2222 to be ENABLED on load balancer 477349
2015/06/24 13:23:19 Creating new node.
POST /v1.0/864477/loadbalancers/477349/nodes HTTP/1.1
Accept: application/json
Content-Type: application/json
User-Agent: gophercloud/1.0.0
X-Auth-Token: ...

{
  "nodes": [
    {
      "address": "10.209.224.145",
      "condition": "ENABLED",
      "port": 2222
    }
  ]
}
HTTP/1.1 422 Unprocessable Entity
Content-Length: 109
Content-Type: application/json
Date: Wed, 24 Jun 2015 13:23:19 GMT
Server: Jetty(8.0.y.z-SNAPSHOT)
Via: 1.1 Rackspace Cloud Load Balancer API v1.24.1 (Repose/2.11.0)
{
  "code": 422,
  "message": "Load Balancer '477349' has a status of 'PENDING_UPDATE' and is considered immutable."
}
2015/06/24 13:23:19 Something went terribly wrong during node creation: []nodes.Node(nil)
```

### DIFF
--- a/pagination/http.go
+++ b/pagination/http.go
@@ -36,13 +36,19 @@ func PageResultFrom(resp *http.Response) (PageResult, error) {
 		parsedBody = rawBody
 	}
 
+	return PageResultFromParsed(resp, parsedBody), err
+}
+
+// PageResultFromParsed constructs a PageResult from an HTTP response that has already had its
+// body parsed as JSON (and closed).
+func PageResultFromParsed(resp *http.Response, body interface{}) PageResult {
 	return PageResult{
 		Result: gophercloud.Result{
-			Body:   parsedBody,
+			Body:   body,
 			Header: resp.Header,
 		},
 		URL: *resp.Request.URL,
-	}, err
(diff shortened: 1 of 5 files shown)


smashwilson:
And now this is unearthing another, previously hidden error in `TestCreate`:

```
--- FAIL: TestCreate (0.00s)
	convenience.go:34: Failure in requests_test.go, line 85: unexpected error "http: read on closed response body"
```

:mag_right:

smashwilson:
:shipit:

PR #439, opened 2015-06-29
IDFromName for networking resources

jrperritt:


jrperritt:
These are nearly exactly like the ones for the `compute` resources :shipit:

PR #440, opened 2015-06-30
IDFromName functions for block storage

jrperritt:


jrperritt:
more IDFromName  functions.

PR #441, opened 2015-07-02
[wip] Logging

jrperritt:


openstack/client.go
```diff
@@ -31,17 +32,21 @@ func NewClient(endpoint string) (*gophercloud.ProviderClient, error) {
 	endpoint = gophercloud.NormalizeURL(endpoint)
 	base = gophercloud.NormalizeURL(base)
 
+	pc := &gophercloud.ProviderClient{
+		IdentityBase: base,
+		Logger: &logrus.Logger{
```
jrperritt: Should we (as above) initialize a logger for users who create a client or let them create it themselves?

jrperritt:
I'm closing this in favor of letting users specify an `HTTPClient` in the `ProviderClient`. That'll let everyone log whatever they want.

jrperritt:
See here for an example: https://github.com/smashwilson/loghttp/blob/master/client.go

PR #442, opened 2015-07-08
Add convenience fns for checking errors

jamiehannaford:


jamiehannaford:
actually, on second thought, I think sticking to `AssertEquals` is a lot simpler

PR #444, opened 2015-07-13
Add MD5 checksum check to file uploads

jamiehannaford:


openstack/objectstorage/v1/objects/requests.go
```diff
@@ -214,11 +216,30 @@ func Create(c *gophercloud.ServiceClient, containerName, objectName string, cont
...
+		return resp, err
 	}
-	res.Err = err
+
+	hash := md5.New()
+	io.Copy(hash, content)
+	localChecksum := hash.Sum(nil)
+
+	for i := 1; i <= 3; i++ {
+		resp, err := doUpload()
```
jrperritt: Pure curiosity: any reason to have the `doUpload` function instead of putting the 4 lines in the `for`-loop?

jrperritt:
Looks good. Maybe a comment above the `Create` function mentioning that the upload gets tried up to 3 times.

openstack/objectstorage/v1/objects/requests.go
jamiehannaford: no - now that I look again, there's no major reason apart from separating things a bit more. do you have a preference? I can move to the main loop if so

jrperritt: Nope, no preference here :) No action required

jrperritt:
+2

PR #446, opened 2015-07-16
Rackspace Flavor extra_specs field

jrperritt:
This PR makes the `extra_specs` field available in the Rackspace `flavors.Flavor` `struct`.


jamiehannaford:
LGTM

PR #447, opened 2015-07-22
Add all_tenants to nova servers list

DSpeichert:
"nova list" functionality in gophercloud is missing an optional parameter "all_tenants" which is useful for users with admin role. Compatible (ignored) for non-admins.


jrperritt:
LGTM

PR #448, opened 2015-07-22
[RFR] OpenStack Telemetry v2

DSpeichert:
Adding support for OpenStack Telemetry v2 API (Ceilometer).

Basic functions are working. Need to add unit/acceptance tests.

### DIFF
--- a/openstack/telemetry/v2/meters/fixtures.go
+++ b/openstack/telemetry/v2/meters/fixtures.go
@@ -4,7 +4,6 @@ package meters
 
 import (
 	"fmt"
-	"log"
 	"net/http"
 	"testing"
 
@@ -16,7 +15,7 @@ import (
 const MeterListBody = `
 [
     {
-        "meter_id": "YmQ5NDMxYzEtOGQ2OS00YWQzLTgwM2EtOGQ0YTZiODlmZDM2K2luc3RhbmNl\n",
+        "meter_id": "YmQ5NDMxYzEtOGQ2OS00YWQzLTgwM2EtOGQ0YTZiODlmZDM2K2luc3RhbmNl",
         "name": "instance",
         "project_id": "35b17138-b364-4e6a-a131-8f3099c5be68",
         "resource_id": "bd9431c1-8d69-4ad3-803a-8d4a6b89fd36",
@@ -30,10 +29,10 @@ const MeterListBody = `
 		"name": "cpu_util",
 		"resource_id": "5b88239b-8ba1-44ff-a154-978cbda23479",
 		"source": "region2",
-		"meter_id": "NWI4ODIzOWAtOGJhMS00NGZhLWExNTQtOTc4Y2JkYTIzNDc5K2NwdV91dGls\n",
+		"meter_id": "NWI4ODIzOWAtOGJhMS00NGZhLWExNTQtOTc4Y2JkYTIzNDc5K2NwdV91dGls",
 		"project_id": "69e6e7c4ed8b434e92feacbf3d4891fd",
 		"type": "gauge",
-		"unit": "%"
+		"unit": "%%"
 	}
 ]
 `
@@ -90,7 +89,7 @@ const MeterStatisticsBody = `
         "period_start": "2015-07-17T16:43:31",
         "avg": 1.5856751105737468,
         "groupby": null,
-        "unit": "%"
+        "unit": "%%"
     }
 ]
 `
@@ -136,7 +135,7 @@ var (
 		Unit:          "GiB",
 	}
 
-	// StatisticsDerp is a MeterStatisticsResult struct that should correspond to the first result in *[]MeterStatisticsResult.
+	// StatisticsDerp is a MeterStatisticsResult struct that should correspond to the second result in *[]MeterStatisticsResult.
 	StatisticsDerp = MeterStatisticsResult{
 		Avg:           1.5856751105737468,
 		Count:         28162,
@@ -186,6 +185,5 @@ func HandleMeterStatisticsSuccessfully(t *testing.T) {
 		w.Header().Add("Content-Type", "application/json")
 		w.WriteHeader(http.StatusOK)
 		fmt.Fprintf(w, MeterStatisticsBody)
-		log.Printf("%s\n", MeterStatisticsBody)
 	})
 }
--- a/openstack/telemetry/v2/meters/fixtures.go
+++ b/openstack/telemetry/v2/meters/fixtures.go
@@ -1,0 +1,191 @@
+//// +build fixtures
+
+package meters
+
+import (
+	"fmt"
+	"log"
+	"net/http"
+	"testing"
+
+	th "github.com/rackspace/gophercloud/testhelper"
+	"github.com/rackspace/gophercloud/testhelper/client"
+)
+
+// MeterListBody contains the canned body of a meters.List response.
+const MeterListBody = `
+[
+    {
+        "meter_id": "YmQ5NDMxYzEtOGQ2OS00YWQzLTgwM2EtOGQ0YTZiODlmZDM2K2luc3RhbmNl\n",
+        "name": "instance",
+        "project_id": "35b17138-b364-4e6a-a131-8f3099c5be68",
+        "resource_id": "bd9431c1-8d69-4ad3-803a-8d4a6b89fd36",
+        "source": "openstack",
+        "type": "gauge",
+        "unit": "instance",
+        "user_id": "efd87807-12d2-4b38-9c70-5f5c2ac427ff"
+    },
+	{
+		"user_id": "7ya0f7a33717400b951037d55b929c53",
+		"name": "cpu_util",
+		"resource_id": "5b88239b-8ba1-44ff-a154-978cbda23479",
+		"source": "region2",
+		"meter_id": "NWI4ODIzOWAtOGJhMS00NGZhLWExNTQtOTc4Y2JkYTIzNDc5K2NwdV91dGls\n",
+		"project_id": "69e6e7c4ed8b434e92feacbf3d4891fd",
+		"type": "gauge",
+		"unit": "%"
+	}
+]
+`
+
+// MeterShowBody is the canned body of a Get request on an existing meter.
+const MeterShowBody = `
+[
+    {
+        "counter_name": "instance",
+        "counter_type": "gauge",
+        "counter_unit": "instance",
+        "counter_volume": 1.0,
+        "message_id": "5460acce-4fd6-480d-ab18-9735ec7b1996",
+        "project_id": "35b17138-b364-4e6a-a131-8f3099c5be68",
+        "resource_id": "bd9431c1-8d69-4ad3-803a-8d4a6b89fd36",
+        "resource_metadata": {
+            "name1": "value1",
+            "name2": "value2"
+        },
+        "source": "openstack",
+        "timestamp": "2013-11-21T12:33:08.323533",
+        "user_id": "efd87807-12d2-4b38-9c70-5f5c2ac427ff"
+    }
+]
+`
+
+// MeterStatisticsBody is the canned body of a Get statistics request on an existing meter.
+const MeterStatisticsBody = `
+[
+    {
+        "avg": 4.5,
+        "count": 10,
+        "duration": 300.0,
+        "duration_end": "2013-01-04T16:47:00",
+        "duration_start": "2013-01-04T16:42:00",
+        "max": 9.0,
+        "min": 1.0,
+        "period": 7200,
+        "period_end": "2013-01-04T18:00:00",
+        "period_start": "2013-01-04T16:00:00",
+        "sum": 45.0,
+        "unit": "GiB"
+    },
+    {
+        "count": 28162,
+        "duration_start": "2015-06-27T20:52:08",
+        "min": 0.06999999999999999,
+        "max": 10.06799336650083,
+        "duration_end": "2015-07-27T20:47:02",
+        "period": 0,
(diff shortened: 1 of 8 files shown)


DSpeichert:
Can somebody more experienced with the test framework (test client) here check why tests fail in `/src/github.com/rackspace/gophercloud/openstack/telemetry/v2/meters` (specifically `requests_test.go`)?

The code works properly against live API server and I went on to copy the output from API server into `/src/github.com/rackspace/gophercloud/openstack/telemetry/v2/meters/fixtures.go`, however the client does don't "discover" anything out of the response body. In https://github.com/DSpeichert/gophercloud/blob/bf1a572b154fba1b4164b353b085df10440c153b/openstack/telemetry/v2/meters/requests.go#L89-L90, the structure while running test looks like this:
```
{Result:{Body:<nil> Header:map[] Err:<nil>}}
&{Status:200 OK StatusCode:200 Proto:HTTP/1.1 ProtoMajor:1 ProtoMinor:1 Header:map[Content-Type:[application/json] Date:[Mon, 03 Aug 2015 18:41:40 GMT] Content-Length:[827]] Body:0xc208034240 ContentLength:827 TransferEncoding:[] Close:false Trailer:map[] Request:0xc20801e8f0 TLS:<nil>}
```

Since the same code works properly with live API, I'm assuming I haven't properly coded the test. Can anybody help?

jrperritt:
A few notes after looking at your code briefly:

1. We don't use named returns in this library. It's certainly subjective, but we prefer to be explicit about initializing and returning variables.

2. You don't need to have pointers to slices. Slices are already pointers under the hood. [Here](https://github.com/DSpeichert/gophercloud/blob/bf1a572b154fba1b4164b353b085df10440c153b/openstack/telemetry/v2/meters/results.go#L26), for example, a regular slice will do.

3. The functions in `request.go` should only return a single type (that inherits `gophercloud.Result`). See [here](https://github.com/DSpeichert/gophercloud/blob/bf1a572b154fba1b4164b353b085df10440c153b/openstack/telemetry/v2/meters/requests.go#L78). The error should be set on the response type and checked in the `Extract` method.

4. Your code in `fixtures.go` looks fine. The result from the test in your output above looks valid (ContentLength = 827). To see if it's getting decoded properly in the `Extract` method,  you can add an output line like `fmt.Printf("response: %+v\n", response)` somewhere like [here](https://github.com/DSpeichert/gophercloud/blob/bf1a572b154fba1b4164b353b085df10440c153b/openstack/telemetry/v2/meters/results.go#L46).

DSpeichert:
Thank you for the comments, @jrperritt.
When it comes to (4), the problem is that the Extract method doesn't get to decode anything. In the output above, the first line shows that Body is nil for some unknown reason (that's output from within Extract function). This is specific to the test helpers. Any idea why it doesn't decode the generic reply at all?

dudymas:
See my PR to your branch, the issue was around the % in the test data

In the handler, you call Fprintf with a string that has a unescaped % ... rather than panicking, the code keeps running and Fprintf just inlines the error. Of course, json can't parse this, but somehow the error was lost (that is new to me. looks like it should have bubbled up). Anyways... I went ahead and corrected that and the newlines in the test data.

DSpeichert:
@jrperritt Please review again.

DSpeichert:
@jrperritt Let me know if you'd like the commits squashed.

jrperritt:
No, the commits are fine, I'm just short on time this week to review PRs.

openstack/telemetry/v2/meters/requests.go
```diff
@@ -0,0 +1,92 @@
...
+// ListOptsBuilder allows extensions to add additional parameters to the
+// List request.
+type ListOptsBuilder interface {
+	ToMeterListQuery() (string, error)
+}
+
+// ListOpts allows the filtering and sorting of collections through
+// the API. Filtering is achieved by passing in struct field values that map to
+// the server attributes you want to see returned.
+type ListOpts struct {
```
jrperritt: Are there no fields? Looks like there should at least be `limit` and `marker`.

openstack/telemetry/v2/meters/requests.go
```diff
@@ -0,0 +1,92 @@
...
+func (opts ListOpts) ToMeterListQuery() (string, error) {
+	q, err := gophercloud.BuildQueryString(opts)
+	if err != nil {
+		return "", err
+	}
+	return q.String(), nil
+}
+
+// List makes a request against the API to list meters accessible to you.
+func List(client *gophercloud.ServiceClient, opts ListOptsBuilder) listResult {
```
jrperritt: All `List` operations should return a `pagination.Pager` like here: https://github.com/rackspace/gophercloud/blob/master/openstack/compute/v2/servers/requests.go#L67

openstack/telemetry/v2/meters/requests.go
```diff
@@ -0,0 +1,92 @@
...
+func (opts MeterStatisticsOpts) ToMeterStatisticsQuery() (string, error) {
+	q, err := gophercloud.BuildQueryString(opts)
+	if err != nil {
+		return "", err
+	}
+	return q.String(), nil
+}
+
+// List makes a request against the API to list meters accessible to you.
+func MeterStatistics(client *gophercloud.ServiceClient, n string, opts MeterStatisticsOptsBuilder) statisticsResult {
```
jrperritt: This looks like another pager with `marker` and `limit`: http://developer.openstack.org/api-ref-telemetry-v2.html#listSamplesStatistics

openstack/telemetry/v2/meters/results.go
```diff
@@ -0,0 +1,99 @@
+package meters
+
+import (
+	"reflect"
+
+	"github.com/mitchellh/mapstructure"
+	"github.com/rackspace/gophercloud"
+)
+
+type MeterListResult struct {
```
jrperritt: Since we're in the `meters` package, just `ListResult`.

openstack/telemetry/v2/meters/results.go
```diff
@@ -0,0 +1,99 @@
...
+	Name       string `json:"name"`
+	ProjectId  string `mapstructure:"project_id"`
+	ResourceId string `mapstructure:"resource_id"`
+	Source     string `json:"source"`
+	Type       string `json:"type"`
+	Unit       string `json:"user"`
+	UserId     string `mapstructure:"user_id"`
+}
+
+type listResult struct {
```
jrperritt: Oh, it looks like the above was caused by a naming conflict.

openstack/telemetry/v2/meters/results.go
jrperritt: In light of `listResult` below, how about just `MeterResult`? Unless there are different `struct` formats for different meter operations?

openstack/telemetry/v2/meters/results.go
```diff
@@ -0,0 +1,99 @@
...
+
+	err = decoder.Decode(r.Body)
+	if err != nil {
+		return nil, err
+	}
+
+	return response, nil
+}
+
+type MeterStatisticsResult struct {
```
jrperritt: Same as above: maybe just `Statistics`. Without context, it's vague, but in context it would be called like `meters.Statistics`, which makes perfect sense.

openstack/telemetry/v2/meters/requests.go
```diff
@@ -0,0 +1,92 @@
...
+// StatisticsOpts allows the filtering and sorting of collections through
+// the API. Filtering is achieved by passing in struct field values that map to
+// the server attributes you want to see returned.
+type MeterStatisticsOpts struct {
+	QueryField string `q:"q.field"`
+	QueryOp    string `q:"q.op"`
+	QueryValue string `q:"q.value"`
+
+	// Optional group by
+	GroupBy string `q:"groupby"`
```
jrperritt: Does this parameter allow multiple strings? If so, it should probably be a `[]string`, and the `strings.Join` in the `ToMeterStatisticsQuery`.

jrperritt:
@DSpeichert Looking good. Just a few changes for consistency with the other services.

openstack/telemetry/v2/meters/results.go
jrperritt: Also, this should be `ListResult` (capitalized).

openstack/telemetry/v2/meters/results.go
```diff
@@ -0,0 +1,99 @@
...
+	Max           float32 `json:"max"`
+	Min           float32 `json:"min"`
+	Period        int     `json:"user_id"`
+	PeriodEnd     string  `mapstructure:"period_end"`
+	PeriodStart   string  `mapstructure:"period_start"`
+	Sum           float32 `json:"sum"`
+	Unit          string  `json:"unit"`
+}
+
+type statisticsResult struct {
```
jrperritt: Same as above: `StatisticsResult`

openstack/telemetry/v2/meters/requests.go
DSpeichert: I believe it allows multiple pairs of "q", which would be pairs of all three.
I couldn't find any example or documentation of how to send a request with multiple "q"s, so I limited the ability to just sending one query.

openstack/telemetry/v2/meters/requests.go
DSpeichert: It would be a page, but the API response body does not provide a link to next page, therefore it would be impossible to code ```NextPageURL()``` function and in my understanding, paging wouldn't work automagically on gophercloud's side. How should I solve that?

carlpett:
@DSpeichert @jrperritt What's the status of this PR? Anything that I can help with to get it done?

DSpeichert:
@carlpett Good question. I have not had time to look into it for a while. It looks like just some of the structs need to be renamed per comments. I'll try to get on that soon.

carlpett:
@DSpeichert I made a PR on your PR-branch, updating the naming and also adding the Show-operation, which I needed.

DSpeichert:
@jrperritt Please review again.

carlpett:
@DSpeichert I did some digging, turns out that you can send a list of queries by simply repeating the parameters (for instance `q.field=timestamp&q.op=gt&q.value=2016-06-06T08:00:00&q.field=message_id&q.op=eq&q.value=ffa19eee-2bae-11e6-afbc-525400b6bb71`). I'm a bit pressed for time at the moment, but it seems to me it should probably be included before merging since it will likely break the contract otherwise? If you do not have time either I'll see if I can get it done during the weekend or so.

carlpett:
Ok, I had a "how hard could it be" moment and started looking at this. Looks like it'd require quite large changes to `BuildQueryString`, however. 
Unless doing something hacky-smelling... For the use-cases I've got at least, it works to append the results of multiple calls to `BuildQueryString`. But there might be some cases where it does not, and it _does_ smell a bit.

So, maybe skip it for now?

DSpeichert:
I don't recall the implementation details but allowing for repeated query parameters would require some reorganization in the code and I'm not sure if it's necessary in the first version. Typically, sending one query at a time seems sufficient - at least it was for me. If you have time, please suggest the improvement.

carlpett:
No, I agree, after seeing the amount of work likely needed (unless doing the multiple `BuildQueryString` thing I mentioned, which I don't think is a good idea), I don't think it is worth it either.

openstack/telemetry/v2/meters/requests.go
jrperritt: What about this one?

DSpeichert: @jrperritt Added supported fields here.

openstack/telemetry/v2/meters/fixtures.go
```diff
@@ -0,0 +1,206 @@
+//// +build fixtures
```
carlpett: @DSpeichert I thought I had done a PR on your branch on this line, but it seems I imagined it, can't find even find it on my own repo any more... The "double-comment" causes this to always be built, and since it imports `testing`, anything built against it will be identified as a test build (and `flag` will add a bunch of testing flags, which is how I noticed)

DSpeichert: Good catch, fixed!

jrperritt:
Ack. Sorry about that. It must be `client.go`. If you rebase on `master` I'll merge this.

DSpeichert:
@jrperritt Rebased.

jrperritt:
Thank you. +2

codevulture:
@jrperritt , @DSpeichert , @dudymas  : I got an error while updating dependency while following https://github.com/kubernetes/kubernetes/blob/master/docs/devel/godep.md , and i think this is related to this change if my env is not corrupt,...Can you please  let me know on how to proceed?

Error:
vagrant@vagrant-ubuntu-trusty-64:~/kubernetes/kubernetes_commit/kuber_play/src/k8s.io/kubernetes$ godep get github.com/rackspace/gophercloud/...
godep: [WARNING]: godep should only be used inside a valid go package directory and
godep: [WARNING]: may not function correctly. You are probably outside of your $GOPATH.
godep: [WARNING]:	Current Directory: /home/vagrant/kubernetes
godep: [WARNING]:	$GOPATH: /home/vagrant/kubernetes/kubernetes_commit/kuber_play
# github.com/rackspace/gophercloud/openstack/telemetry/v2/meters
kubernetes_commit/kuber_play/src/github.com/rackspace/gophercloud/openstack/telemetry/v2/meters/results.go:78: cannot use decoderHooks (type func(reflect.Type, reflect.Type, interface {}) (interface {}, error)) as type mapstructure.DecodeHookFunc in field value
godep: go exit status 2
godep: exit status 1

PR #451, opened 2015-07-29
allow token authentication

jrperritt:


smashwilson:
> nit: since we have UserID, DomainID and TenantID it would probably make more sense to have TokenID

Hmm. For those three, the value is an identifier rather than the thing itself: a numeric value that's used to look up the user, domain or tenant in a database somewhere, rather than the user, domain or tenant object itself. But to me, the token _is_ the token, so calling it `Token` without the `ID` makes sense.

Naming things!

openstack/identity/v2/tokens/requests.go
```diff
@@ -38,20 +42,24 @@ func (auth AuthOptions) ToTokenCreateMap() (map[string]interface{}, error) {
...
+			}
+		} else {
+			return nil, ErrPasswordRequired
+		}
+	} else if auth.Token != "" {
+		authMap["token"] = map[string]interface{}{
+			"id": auth.Token,
+		}
+	} else {
+		return nil, fmt.Errorf("You must provide either username/password or tenantID/token values.")
```
smashwilson: Could we keep this another error constant, like the other return values? It makes it easier for clients to test against:

```go
_, err := tokens.Create()
if err == tokens.ErrNoUsernameOrToken {
  // ..
}
```

jamiehannaford:
@smashwilson But isn't the `Token` field a [string representation](https://github.com/rackspace/gophercloud/pull/451/files#diff-8164891aafc5e03b485c8663f27d8f5eR49), i.e. an identifier only.

openstack/identity/v2/tokens/requests.go
jamiehannaford: do we need a specific error for this or could we use `InvalidInputError` instead? it seems like an extremely specific error name, when really it's just invalid input. plus if we're being accurate it would really be `ErrNoUsernameOrPasswordOrTokenOrTenant` :stuck_out_tongue:

smashwilson: Haha, very true. `InvalidInputError` would be perfectly fine. Actually, once #452 makes it in, we could scrub all of these errors out in favor of the better, generic ones - although that would be a backwards-incompatible change...

jrperritt: That's what I'm thinking. Any errors I've added here are purely place-holders until we can change it to an error type.

jrperritt:
https://github.com/rackspace/gophercloud/commit/1f218c8cf1cd9622a15474824a442c3b396ee926 is what the field would look like as as `TokenID`. I'm torn here because what OpenStack calls the token ID is actually the token, and what it calls a token is an object with other information like the expiration date. I'm indifferent here.

smashwilson:
I'd vote for consistency with OpenStack terminology then, which it [looks like](http://developer.openstack.org/api-ref-identity-v2.html#authenticate-v2.0) is `TokenID`. :+1:

PR #452, opened 2015-07-29
[wip] Error types

jrperritt:


errors.go
```diff
@@ -0,0 +1,50 @@
...
+	return "Authentication failed"
+}
+func (e defaultError404) Error() string {
+	return "Resource not found"
+}
+func (e defaultError405) Error() string {
+	return "Method not allowed"
+}
+
+// Error401er is the interface resource implement to override the error message
```
smashwilson: Sounds like this comment is missing a word somewhere. "Error401er is the interface resource [..] implement..." Maybe "... is the interface resource _error types_ implement to override..."?

Ditto for the other two Error interfaces below.

provider_client.go
```diff
@@ -215,13 +200,47 @@ func (client *ProviderClient) Request(method, url string, options RequestOpts) (
...
+				}
+				if options.RawBody != nil {
+					options.RawBody.Seek(0, 0)
+				}
+				resp, err = client.Request(method, url, options)
+				if err != nil {
+					return nil, fmt.Errorf("Successfully re-authenticated, but got error executing request: %s", err)
+				}
+			}
+			err = defaultError401{}
```
smashwilson: Should these error constructors include the `UnexpectedResponseCodeError`?

```go
err = defaultError401{respErr}
```

provider_client.go
```diff
@@ -215,13 +200,47 @@ func (client *ProviderClient) Request(method, url string, options RequestOpts) (
...
+		case http.StatusNotFound:
+			err = defaultError404{}
+			if _, ok := errType.(Error404er); ok {
+				err = errType.(Error404er).Error404(respErr)
+			}
+		case http.StatusMethodNotAllowed:
+			err = defaultError405{}
+			if _, ok := errType.(Error405er); ok {
+				err = errType.(Error405er).Error405(respErr)
+			}
```
smashwilson: :+1: Not bad, not bad.

I think you could save yourself the double cast like so:

```go
if err405, ok := errType.(Error405er); ok {
  err = err405.Error405(respErr)
}
```

provider_client.go
```diff
@@ -88,18 +88,19 @@ type RequestOpts struct {
...
 	// the response has a different code, an error will be returned.
 	OkCodes []int
-
 	// MoreHeaders specifies additional HTTP headers to be provide on the request. If a header is
 	// provided with a blank value (""), that header will be *omitted* instead: use this to suppress
 	// the default Accept header or an inferred Content-Type, for example.
 	MoreHeaders map[string]string
+	// ErrorType specifies the resource error type to return if an error is encountered.
+	// This lets resources override default error messages based on the response status code.
+	ErrorType error
```
smashwilson: Another naming nit: it's not really a "type" here, since you're passing a preconstructed error template of sorts. `ErrorContext` maybe, because it's really used to capture request values that can be used to build the eventual error? `ErrorTemplate`? `ErrorExemplar`?

smashwilson:
This is off to a good start :+1:

I agree with @jamiehannaford about splitting an `InvalidInputError` out of `BaseError`, or even just renaming `BaseError` to `InvalidInputError` if it's the only one right now.

One consideration to keep in mind is giving users an ergonomic way to programmatically test for different kinds of errors without having to use regular expressions against the message or some such. There are a few ways that we could go from here to enable this and I'm not sure which one I prefer. To brainstorm a little:

### Test methods on gophercloud error types (1)

Allow users to call methods on the errors to distinguish them:

```go
provider, err := rackspace.AuthenticatedClient(opts)
if err != nil {
  if err.IsAuthFailure() {
    log.Fatalf("You gave me bad credentials!")
  }
}
```

*To make this work:*

1. "Inherit" `InvalidInputError` and `UnexpectedResponseCodeError` from a common error type. Maybe that's `BaseError`?
2. Define all possible `IsXyz()` methods on `BaseError` to return false.
3. Override the appropriate `IsXyz()` methods to return true on error subtypes. `UnexpectedResponseCodeError` can test its response code and return true if its code is a 401, for example.
4. Return `gophercloud.BaseError` instead of `err` from basically everything. (This is the painful bit here.)

### Hide the cast and error check in utility functions (2)

Provide utility functions to test any `error` against a condition:

```go
provider, err := rackspace.AuthenticatedClient(opts)
if gophercloud.IsAuthFailure(err) {
  log.Fatalf("It looks like they 'fixed the glitch'")
}
```

*To make this work:*

1. Write those utility functions. Accept any `error` type, but test for `nil` and attempt to cast against the appropriate error.

The tricky part here is getting those casts right. You [can't cast to an embedded type](http://play.golang.org/p/Up2C7ehnLC), so you'd need to cast-test against *each* error type, somehow. There may be some solution with a common parent interface shared by all gophercloud errors but I don't think I quite know how the specifics of that would look.

### Error codes (3)

Include a "Code" field in gophercloud errors that contains some exported constant you can check against.

```go
provider, err := rackspace.AuthenticatedClient(opts)
if err != nil {
  if err.Code == gophercloud.ErrAuthFailure {
    log.Fatalf("I don't even know who you are anymore")
  }
}
```

*To make this work:*

1. We'll need that common base error type again, to include the `Code` field and return from methods.
2. Export a bunch of `const` fields that you can check against.

errors.go
jrperritt: I'd originally intended resource to be plural, but you're right: it's the error types that should be mentioned.

provider_client.go
jrperritt: We could do that. Currently it's being passed to the `ErrorXXX` method, but it's true that it should probably also go to the more general resource error method.

provider_client.go
jrperritt: :+1:

provider_client.go
smashwilson: Oh, I'd just noticed that the `defaultError401` type definitions all embed an `*UnexpectedResponseCodeError`:

```go
type defaultError405 struct {
	*UnexpectedResponseCodeError
}
```

provider_client.go
jrperritt: Good point. `ErrorContext` sounds better.

jrperritt:
>> One consideration to keep in mind is giving users an ergonomic way to programmatically test for different kinds of errors without having to use regular expressions against the message or some such.

One idea I had for this is adding a `Type` field to the `InvalidInputError` and let users `switch` on that:
```
server, err := servers.Get(client, id).Extract()
if err != nil {
  if invalidInputError, ok := err.(gophercloud.InvalidInputError); ok {
    switch invalidInputError.Type {
      case "MissingIDField":
      case "MissingUsername":
    }
  }
}
```

jrperritt:
Closing this since it'll be backwards incompatible here. It'll go in `gophercloud/gophercloud`.

PR #453, opened 2015-07-30
Object create: ETag bug fix

jrperritt:
Once the `io.Reader` is read, it can't be re-read. This PR simultaneously creates a hash of the object and writes the content to a buffer. The buffer is then converted to a reader and passed as the request body.


jrperritt:
Verified working locally. :shipit:  because bug fix

PR #456, opened 2015-08-04
add DeleteMetadata field for deleting container metadata

jrperritt:


pratikmallya:
unit tests would be a nice addition

jrperritt:
I agree.

PR #457, opened 2015-08-05
Add keystone v3 projects basic CRUD

kwapik:


kwapik:
According to [this file](https://github.com/rackspace/gophercloud/blob/master/CONTRIBUTING.md), pull requests should be made to development branch (i.e. `release/v1.0.1`), but after looking into commit history I realized that it does not make sense. 
So:
* should I change my PR to fit master branch (it seems to be active)?
* why CONTRIBUTING.md is outdated?

PR #461, opened 2015-08-11
Add more APIs for Cinder

feiskyer:
Add more APIs for Cinder:

* attach
* detach
* reserve
* unreserve
* initialize_connection
* terminate_connection

### DIFF
--- a/openstack/blockstorage/v1/volumes/requests.go
+++ b/openstack/blockstorage/v1/volumes/requests.go
@@ -234,3 +234,170 @@ func IDFromName(client *gophercloud.ServiceClient, name string) (string, error) 
 		return "", fmt.Errorf("Found %d volumes matching %s", volumeCount, name)
 	}
 }
+
+type AttachOptsBuilder interface {
+	ToVolumeAttachMap() (map[string]interface{}, error)
+}
+
+type AttachOpts struct {
+	MountPoint 		string
+	InstanceUUID 	string
+	HostName		string
+	Mode			string
+}
+
+func (opts AttachOpts) ToVolumeAttachMap() (map[string]interface{}, error) {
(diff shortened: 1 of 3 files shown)


feiskyer:
cc @jrperritt @smashwilson   

Will someone merge this PR?

jamiehannaford:
@feiskyer Don't these operations belong in [Cinder v2](http://developer.openstack.org/api-ref-blockstorage-v2.html)? You've added them to the v1 package.

feiskyer:
@jamiehannaford  Those APIs are not list at [Cinder v2](http://developer.openstack.org/api-ref-blockstorage-v2.html) and [Cinder v1](http://developer.openstack.org/api-ref-blockstorage-v1.html), but they are both working.

jrperritt:
Where are these actions documented? I see the `attach` action [here](http://developer.openstack.org/api-ref-blockstorage-v2.html#volume-api-v2-volumes-actions) but not the rest.

feiskyer:
@jrperritt There actions are defined at [here](https://github.com/openstack/cinder/blob/master/cinder/api/contrib/volume_actions.py). They are called directly by nova, but unfortunately not documented. These apis are useful for those who don't need nova.

jrperritt:
Ah, OK. At the least, these operations should be moved the the `v2` package, and unit and acceptance tests should be added. Though they may work with `v1`, I'd prefer to add new functionality to the non-deprecated version.

feiskyer:
@jrperritt #485 committed

jrperritt:
Closing in favor of #485 ; I think that PR contains all this and more. @feiskyer let me know if that's not the case.

feiskyer:
@jrperritt #485 contains all this and more, It's OK to close this one.

Issue #463, opened 2015-08-19
Cloud orchestration time format error


pratikmallya:
I get errors such as `error	parsing time "2015-08-17T13:58:39" as "2006-01-02T15:04:05Z07:00": cannot parse "" as "Z07:00"`, which I think is because the format used by heat is different from that used by time.RFC3339

jrperritt:
I think this was addressed by #464

PR #464, opened 2015-08-19
Fix time format used for cloud orchestration

pratikmallya:
The format used to represent time is slightly different than RFC
standard and the one present in GO time lib.

Fixes https://github.com/rackspace/gophercloud/issues/463

### DIFF
--- a/openstack/orchestration/v1/stackevents/fixtures.go
+++ b/openstack/orchestration/v1/stackevents/fixtures.go
@@ -4,7 +4,6 @@ import (
 	"fmt"
 	"net/http"
 	"testing"
-	"time"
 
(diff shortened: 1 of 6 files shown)


pratikmallya:
The failing tests seem to not be easy to fix just yet. I will take a look at them later

jrperritt:
Thank you for the PR :)

Yes, I think it's just places like [here](https://github.com/rackspace/gophercloud/blob/master/openstack/orchestration/v1/stackevents/fixtures.go#L70) in the unit tests.

pratikmallya:
closed in favor of https://github.com/rackspace/gophercloud/pull/478

PR #466, opened 2015-08-21
Fix Pagination for list StackResources

pratikmallya:
The api does not support marker based pagination and this breaks
the clients which loop infinitely over list resource calls.


pratikmallya:
Might need to make more changes to `ListOpts` and/or remove it entirely.

jrperritt:
+2

PR #467, opened 2015-08-24
Fix validation requests to use correct format for template

pratikmallya:
The orchestration api accepts a template for validation in string
format. This fix aligns the request structure with that
expectation.


pratikmallya:
closing in favor of https://github.com/rackspace/gophercloud/pull/483

PR #468, opened 2015-08-27
[rfr] Keystone Identity /v3/role_assignments

DSpeichert:
This implements the /v3/role_assignments part of roles API, which complements existing v2 functions already implemented in gophercloud. Querying role assignments is, however, only available in v3 and this patch makes it usable.

### DIFF
--- a/openstack/identity/v3/roles/requests.go
+++ b/openstack/identity/v3/roles/requests.go
@@ -1,0 +1,31 @@
+package roles
+
+import (
+	"github.com/rackspace/gophercloud"
+	"github.com/rackspace/gophercloud/pagination"
+)
+
+// RoleAssignmentsOpts allows you to query the RoleAssignments method.
+type RoleAssignmentsOpts struct {
+	GroupId        string `q:"group.id"`
+	RoleId         string `q:"role.id"`
+	ScopeDomainId  string `q:"scope.domain.id"`
+	ScopeProjectId string `q:"scope.project.id"`
+	UserId         string `q:"user.id"`
+	Effective      bool   `q:"effective"`
+}
+
+// RoleAssignments enumerates the roles assigned to a specified resource.
+func RoleAssignments(client *gophercloud.ServiceClient, opts RoleAssignmentsOpts) pagination.Pager {
+	u := roleAssignmentsURL(client)
+	q, err := gophercloud.BuildQueryString(opts)
+	if err != nil {
+		return pagination.Pager{Err: err}
+	}
+	u += q.String()
+	createPage := func(r pagination.PageResult) pagination.Page {
+		return RoleAssignmentsPage{pagination.LinkedPageBase{PageResult: r}}
+	}
+
+	return pagination.NewPager(client, u, createPage)
+}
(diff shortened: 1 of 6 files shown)


DSpeichert:
@jrperritt can you take a look at this one?

jrperritt:
Probably not until next week

openstack/identity/v3/roles/requests.go
```diff
@@ -0,0 +1,31 @@
+package roles
+
+import (
+	"github.com/rackspace/gophercloud"
+	"github.com/rackspace/gophercloud/pagination"
+)
+
+// RoleAssignmentsOpts allows you to query the RoleAssignments method.
+type RoleAssignmentsOpts struct {
```
jrperritt: Could you add some field descriptions in the form of comments?

openstack/identity/v3/roles/requests.go
```diff
@@ -0,0 +1,31 @@
...
+	GroupId        string `q:"group.id"`
+	RoleId         string `q:"role.id"`
+	ScopeDomainId  string `q:"scope.domain.id"`
+	ScopeProjectId string `q:"scope.project.id"`
+	UserId         string `q:"user.id"`
+	Effective      bool   `q:"effective"`
+}
+
+// RoleAssignments enumerates the roles assigned to a specified resource.
+func RoleAssignments(client *gophercloud.ServiceClient, opts RoleAssignmentsOpts) pagination.Pager {
```
jrperritt: This should be named `ListAssignments` for consistency. Also, it should take a `ListAssignmentsOptsBuilder` as a parameter instead of the `RoleAssignmentsOpts` `struct`. You can see an example here: https://github.com/rackspace/gophercloud/blob/master/openstack/networking/v2/networks/requests.go#L32

openstack/identity/v3/roles/urls.go
```diff
@@ -0,0 +1,7 @@
+package roles
+
+import "github.com/rackspace/gophercloud"
+
+func roleAssignmentsURL(client *gophercloud.ServiceClient) string {
```
jrperritt: `listAssignmentsURL`

openstack/identity/v3/roles/results.go
```diff
@@ -0,0 +1,64 @@
+package roles
+
+import (
+	"github.com/rackspace/gophercloud/pagination"
+
+	"github.com/mitchellh/mapstructure"
+)
+
+// RoleAssignment is the result of a role assignments query.
+type RoleAssignment struct {
```
jrperritt: We don't really need the pointer fields here or in `Scope`.

openstack/identity/v3/roles/results.go
```diff
@@ -0,0 +1,64 @@
...
+	ID string `json:"id,omitempty"`
+}
+
+type Group struct {
+	ID string `json:"id,omitempty"`
+}
+
+// RoleAssignmentsPage is a single page of RoleAssignments results.
+type RoleAssignmentsPage struct {
+	pagination.LinkedPageBase
```
jrperritt: If this is a `LinkedPageBase`, you'll need a `NextPageURL` function like here: https://github.com/rackspace/gophercloud/blob/master/openstack/compute/v2/servers/results.go#L189

jrperritt:
Looking good. Also maybe add an acceptance test.

jrperritt:
+2

PR #472, opened 2015-08-28
[RFR] OpenStack Compute Start/Stop Server

DSpeichert:
Added the ability to start/stop OpenStack Nova instances (V2 API Extensions).

Unit tests included.


jrperritt:
Are these the same as this: https://github.com/rackspace/gophercloud/blob/master/openstack/compute/v2/extensions/startstop/requests.go ?

DSpeichert:
OMG. They indeed are. I guess I assumed it wasn't implemented if I didn't see it in the docs. I was wrong...

PR #476, opened 2015-09-08
[rfr] Reorganized volumes and volumeattach to move fixtures to subpackage

### LINKED ISSUE #473: Import of volumes and volumeattach package causes indirect dependency on testing
I guess it is common knowledge that gophercloud is used in kubernetes for openstack

Import of the below packages is resulting in addition of test flags to kubernetes binaries 
```
"github.com/rackspace/gophercloud/openstack/blockstorage/v1/volumes"
"github.com/rackspace/gophercloud/openstack/compute/v2/extensions/volumeattach"
```
If we look in say https://github.com/rackspace/gophercloud/blob/master/openstack/blockstorage/v1/volumes/fixtures.go#L6

We will see that these imports are in turn importing "testing". Which is declaring a bunch of command line args. https://github.com/golang/go/blob/master/src/testing/testing.go#L187

Moving fixtures.go to a testing sub package and updating the dependencies appropriately seems to fix this issue. (thanks to @thockin for the suggestion)
https://github.com/kubernetes/kubernetes/pull/13367

Please let me know if this sounds like a right thing to do, I would be more than happy to send a PR for the same.

spothanis:
This PR is to fix #473. It is basically a re-org of volumes and volumeattach packages to move fixtures.go into its own subpackage. Updated dependencies accordingly.[RFR]

### DIFF
--- a/openstack/blockstorage/v1/volumes/requests_test.go
+++ b/openstack/blockstorage/v1/volumes/requests_test.go
@@ -3,6 +3,7 @@ package volumes
 import (
 	"testing"
 
+	fixtures "github.com/rackspace/gophercloud/openstack/blockstorage/v1/volumes/testing"
 	"github.com/rackspace/gophercloud/pagination"
 	th "github.com/rackspace/gophercloud/testhelper"
 	"github.com/rackspace/gophercloud/testhelper/client"
@@ -12,7 +13,7 @@ func TestList(t *testing.T) {
 	th.SetupHTTP()
 	defer th.TeardownHTTP()
 
-	MockListResponse(t)
+	fixtures.MockListResponse(t)
 
 	count := 0
 
@@ -49,7 +50,7 @@ func TestListAll(t *testing.T) {
 	th.SetupHTTP()
 	defer th.TeardownHTTP()
 
-	MockListResponse(t)
+	fixtures.MockListResponse(t)
 
 	allPages, err := List(client.ServiceClient(), &ListOpts{}).AllPages()
 	th.AssertNoErr(t, err)
@@ -75,7 +76,7 @@ func TestGet(t *testing.T) {
 	th.SetupHTTP()
 	defer th.TeardownHTTP()
 
-	MockGetResponse(t)
+	fixtures.MockGetResponse(t)
 
 	v, err := Get(client.ServiceClient(), "d32019d3-bc6e-4319-9c1d-6722fc136a22").Extract()
 	th.AssertNoErr(t, err)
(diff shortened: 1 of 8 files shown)


eparis:
I know my vote doesn't count for anything, but I like it.

openstack/blockstorage/v1/volumes/requests_test.go
```diff
@@ -3,6 +3,7 @@ package volumes
 import (
 	"testing"
```
saad-ali: This is a test file and imports the `testing` package. It should be moved to the `testing` package as well.

openstack/compute/v2/extensions/volumeattach/requests_test.go
```diff
@@ -3,15 +3,44 @@ package volumeattach
 import (
 	"testing"
```
saad-ali: This one too.

rackspace/blockstorage/v1/volumes/delegate_test.go
```diff
@@ -3,7 +3,8 @@ package volumes
 import (
 	"testing"
```
saad-ali: Ditto

rackspace/compute/v2/volumeattach/delegate_test.go
```diff
@@ -3,24 +3,53 @@ package volumeattach
 import (
 	"testing"
```
saad-ali: Ditto.

openstack/blockstorage/v1/volumes/requests_test.go
saad-ali: Same with `urls_test.go`

jrperritt:
Is this ready for review? Please use `[wip]` and `[rfr]` tags in the issue title so that others know this is either a work in progress or ready for review.

openstack/blockstorage/v1/volumes/requests_test.go
spothanis: @saad-ali I didn't move those cause they were explicitly named test files and were not affecting the production code. I agree that it will probably be cleaner to have all of the test files in the package, however there is some code that is imported within the package, such between urls.go and urls_test.go. So in order to move that it would require altering the scope of the functions in urls.go.

saad-ali: Sounds good, thanks @spothanis

spothanis:
@jrperritt updated the title with RTR.

spothanis:
@jrperritt bump :)

jrperritt:
I'll be able to review later today or tomorrow. Please not though that the request was to put the `[rfr]` in the issue _title_ so that anyone looking at the _Pull requests_ page can see which ones are ready for review. I've amended it for you.

spothanis:
Thanks @jrperritt

jrperritt:
LGTM

+2

spothanis:
Thanks @jrperritt

PR #477, opened 2015-09-09
[wip] ip addresses (neutron extension)

jrperritt:


rackspace/compute/v2/servers/results.go
```diff
@@ -0,0 +1,179 @@
...
+
+	"github.com/mitchellh/mapstructure"
+	"github.com/rackspace/gophercloud"
+	"github.com/rackspace/gophercloud/pagination"
+)
+
+// Server represents a Rackspace server on the user's account.
+type Server struct {
+	// ID uniquely identifies this server amongst all other servers, including those not accessible to the current tenant.
+	ID string
```
jamiehannaford: we can't we embed `os.Server` here? (haven't worked in go in a while :grin: )

rackspace/compute/v2/servers/results.go
```diff
@@ -0,0 +1,179 @@
...
+	serverResult
+}
+
+// RebuildResult temporarily contains the response from a Rebuild call.
+type RebuildResult struct {
+	serverResult
+}
+
+// Extract interprets any serverResult as a Server, if possible.
+func (r serverResult) Extract() (*Server, error) {
```
jamiehannaford: perhaps we can put all of this logic in `results.go` as convention dictates? There's not much code though, so if you think it's overkill your call

rackspace/networking/v2/ipaddresses/requests.go
```diff
@@ -0,0 +1,230 @@
...
+		return nil, fmt.Errorf("Required CreateOpts field 'Version' not set.")
+	}
+	v["version"] = opts.Version
+
+	if opts.TenantID == "" {
+		return nil, fmt.Errorf("Required CreateOpts field 'TenantID' not set.")
+	}
+	v["tenant_id"] = opts.TenantID
+
+	return map[string]interface{}{"ip_address": v}, nil
```
jamiehannaford: :+1: For v2 we could centralise some of this validation logic into helpers

rackspace/networking/v2/ipaddresses/requests.go
```diff
@@ -0,0 +1,230 @@
...
+}
+
+// Create will provision a new shapred IP address based on the values in CreateOpts. To extract
```
jamiehannaford: s/shapred/shared/

rackspace/networking/v2/ipaddresses/requests.go
```diff
@@ -0,0 +1,230 @@
...
+	_, res.Err = client.Post(createURL(client), reqBody, &res.Body, &gophercloud.RequestOpts{
+		OkCodes: []int{201},
+	})
+	return res
+}
+
+// Delete will deallocate the existing shared IP with the provided ID from the tenant.
+// Before using this operation, all IP associations must be removed from the IP address
+// by using the Disassociate function.
+func Delete(client *gophercloud.ServiceClient, id string) DeleteResult {
```
jamiehannaford: ugh... do you think we should do this disassociation for them, or is it better to be explicit and force the user to execute their own destructive actions?

jamiehannaford: maybe we could add a flag?

rackspace/networking/v2/ipaddresses/requests.go
```diff
@@ -0,0 +1,230 @@
...
+func (opts ListOpts) ToSharedIPListByServerQuery() (string, error) {
+	q, err := gophercloud.BuildQueryString(opts)
+	if err != nil {
+		return "", err
+	}
+	return q.String(), nil
+}
+
+// ListByServer returns a Pager which allows you to iterate over a collection of
+// IPAssociations.
```
jamiehannaford: s/IPAssociations/IP addresses associated with the given server/ maybe?

jamiehannaford:
I don't see anything wrong with copy and pasting the Rackspace Server struct into a more distinct thing. We'll likely be adding extra fields in the future, so it's probably best we decouple Rackspace's implementation.

The only question I have is whether _all_ Rackspace API users will have this added field (my guess is yes). If this is not the case, the only other option would be to embed the `os.Server` into a new `NetworkIPServer` struct that lives in the `ipaddresses` subpkg or something, with its own extract fn. I'm not sure if this would even work, but you'd then be isolating this extension from any other.

Also, I know it's a pain, but do we have acceptance tests for this?

rackspace/compute/v2/servers/results.go
jrperritt: I'm not sure I follow: which code should go in `results.go` (this file)?

rackspace/compute/v2/servers/results.go
jrperritt: We _could_, but I don't think we gain anything here by doing it. Specifically, if the OpenStack server has fields that Rackspace doesn't have (now or in the future), we'll end up copy-and-pasting like this anyhow. Only if done that way, it would potentially break code that initializes a Rackspace server (since it'll no longer embed an OpenStack server):
```
server := rsServers.Server{
   osServers.Server{

   },
}
```

rackspace/networking/v2/ipaddresses/requests.go
jrperritt: :+1:

rackspace/networking/v2/ipaddresses/requests.go
jrperritt: :+1:

jrperritt:
Yeah, that field gets added to all of the responses from `servers` that respond with a server :/

Ah, quite right. All that copy-and-paste was to get that field I needed to create the acceptance tests in the first place (to share a public IP, the servers have to have the same value for that field.) I'll add those.

rackspace/networking/v2/ipaddresses/requests.go
jrperritt: Hmm, I think it should be explicit. The flag is a good idea. Let's add a `DeleteOptsBuilder` parameter that we can add fields to later for this. One of the things I'd want in a Gophercloud v2 is for all the operations to take `*OptsBuilder` parameters even if they currently need it so that we can retroactively add parameters like this.

rackspace/compute/v2/servers/results.go
jamiehannaford: me neither that I look at it again. pretty sure this was before my 2nd coffee, please ignore :sunglasses:

rackspace/compute/v2/servers/results.go
jamiehannaford: yeah I agree actually. I think we should decouple the providers as much as possible and not assume openstack is a superset

jrperritt:
Not ready to merge. Permissions issues (likely limited to me) arose while running the acceptance tests.

jrperritt:
Closing. This will go in the Rackspace extension to `gophercloud/gophercloud`

PR #480, opened 2015-09-10
[rfr]Allow DNSNameservers and HostRoutes to be removed

timbyr:
Check against nil instead of len == 0 when updating DNSNameservers and
HostRoutes. This allows the removal of already configured properties.


jrperritt:
+2

PR #481, opened 2015-09-12
[rfr] Allow Empty Image ID and Name

jtopjian:
An empty image ID and Name should be allowable to account for the
boot-from-volume extension.

### DIFF
--- a/openstack/compute/v2/servers/requests.go
+++ b/openstack/compute/v2/servers/requests.go
@@ -260,19 +260,18 @@ func Create(client *gophercloud.ServiceClient, opts CreateOptsBuilder) CreateRes
 		return res
 	}
 
-	// If ImageRef isn't provided, use ImageName to ascertain the image ID.
+	// If ImageRef isn't provided, and ImageName is, use ImageName to ascertain the image ID.
+	// If neither are set, assume boot-from-volume is being used.
 	if reqBody["server"].(map[string]interface{})["imageRef"].(string) == "" {
 		imageName := reqBody["server"].(map[string]interface{})["imageName"].(string)
-		if imageName == "" {
-			res.Err = errors.New("One and only one of ImageRef and ImageName must be provided.")
-			return res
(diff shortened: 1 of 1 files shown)


jtopjian:
@jrperritt I'm trying to sort out some boot-from-volume issues in Terraform. I need to be able to pass an empty Image ID or Name in order to "cleanly" create a volume-booted instance. I'm pretty sure that if an Image ID is still used, the bootable volume takes precedence, but allowing an empty ID and Name may be the proper solution.

jrperritt:
I'm not sure I follow: can you not use the `Create` operation in the `bootfromvolume` package: https://github.com/rackspace/gophercloud/blob/master/openstack/compute/v2/extensions/bootfromvolume/requests.go

jtopjian:
Probably, but I don't think it would be without some major edits to the Terraform `openstack_compute_instance_v2` resource: 

https://github.com/hashicorp/terraform/blob/master/builtin/providers/openstack/resource_openstack_compute_instance_v2.go#L307

jtopjian:
Sorry, I think I understand. The `bootfromvolume` Create just calls the base `Server Create`?

Assuming that's true, wouldn't an ImageRef still be required if the base Create is still being called?

The core issues is that an ImageRef shouldn't be used at all when booting from volume. Note the example given in https://github.com/hashicorp/terraform/issues/3206 : 

```
Attempt to boot from volume - no image supplied
```

jtopjian:
aaand third comment's a charm. Yes, you're right. I re-read both `servers` and `bootfromvolume`, see how they work, and verified as such.

I have some working code in Terraform that I'll cleanup and use.

Thank you for the input. :)

jrperritt:
Ah, I was just typing a response because I realized my first one wasn't very helpful at all. Glad you figured it out though.

jtopjian:
It was helpful as a good push to get me to connect the dots on my own :)

PR #478, opened 2015-09-09
[rfr]Fix time format used for cloud orchestration

pratikmallya:
The format used to represent time is slightly different than RFC
standard and the one present in golang time lib.

PR #483, opened 2015-09-14
[rfr]Fix api interfaces for orchestration resources

pratikmallya:
Some of the interfaces don't correspond well to the values
expected by the requests and returned by api.

depends on: https://github.com/rackspace/gophercloud/pull/478
corresponds to: https://github.com/rackspace/rack/pull/324

### DIFF
--- a/openstack/orchestration/v1/stackresources/results.go
+++ b/openstack/orchestration/v1/stackresources/results.go
@@ -1,6 +1,7 @@
 package stackresources
 
 import (
+	"encoding/json"
 	"fmt"
 	"reflect"
 	"time"
@@ -12,15 +13,18 @@ import (
 
 // Resource represents a stack resource.
 type Resource struct {
-	Links        []gophercloud.Link `mapstructure:"links"`
-	LogicalID    string             `mapstructure:"logical_resource_id"`
-	Name         string             `mapstructure:"resource_name"`
-	PhysicalID   string             `mapstructure:"physical_resource_id"`
-	RequiredBy   []interface{}      `mapstructure:"required_by"`
-	Status       string             `mapstructure:"resource_status"`
-	StatusReason string             `mapstructure:"resource_status_reason"`
-	Type         string             `mapstructure:"resource_type"`
-	UpdatedTime  time.Time          `mapstructure:"-"`
+	Attributes   map[string]interface{} `mapstructure:"attributes"`
+	CreationTime time.Time              `mapstructure:"-"`
+	Description  string                 `mapstructure:"description"`
+	Links        []gophercloud.Link     `mapstructure:"links"`
+	LogicalID    string                 `mapstructure:"logical_resource_id"`
+	Name         string                 `mapstructure:"resource_name"`
+	PhysicalID   string                 `mapstructure:"physical_resource_id"`
+	RequiredBy   []interface{}          `mapstructure:"required_by"`
+	Status       string                 `mapstructure:"resource_status"`
+	StatusReason string                 `mapstructure:"resource_status_reason"`
+	Type         string                 `mapstructure:"resource_type"`
+	UpdatedTime  time.Time              `mapstructure:"-"`
 }
 
 // FindResult represents the result of a Find operation.
@@ -54,6 +58,13 @@ func (r FindResult) Extract() ([]Resource, error) {
 			}
 			res.Res[i].UpdatedTime = t
 		}
+		if date, ok := resource["creation_time"]; ok && date != nil {
+			t, err := time.Parse(gophercloud.STACK_TIME_FMT, date.(string))
+			if err != nil {
+				return nil, err
+			}
+			res.Res[i].CreationTime = t
+		}
 	}
 
 	return res.Res, nil
@@ -75,18 +86,6 @@ func (r ResourcePage) IsEmpty() (bool, error) {
 	return len(resources) == 0, nil
 }
 
-// LastMarker returns the last container name in a ListResult.
-func (r ResourcePage) LastMarker() (string, error) {
-	resources, err := ExtractResources(r)
-	if err != nil {
-		return "", err
-	}
-	if len(resources) == 0 {
-		return "", nil
-	}
-	return resources[len(resources)-1].PhysicalID, nil
-}
-
 // ExtractResources interprets the results of a single page from a List() call, producing a slice of Resource entities.
 func ExtractResources(page pagination.Page) ([]Resource, error) {
 	casted := page.(ResourcePage).Body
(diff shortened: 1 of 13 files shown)


jrperritt:
@pratikmallya Because of the built-in way Go handles package management (via `go get`), it's important that we are aware of backwards-incompatible changes that aren't bugs. For example, the issue of having the wrong time format I'd consider a bug, because users could not have gotten that working with their own code. Also, adding fields and functions/methods is also fine. However, changing types and removing fields is a bit more sensitive. For each of the changes in this PR where you changed a data type or removed fields, can you provide an explanation for why? When the orchestration service was implemented, it was coded against the docs [here](http://developer.openstack.org/api-ref-orchestration-v1.html) and [here](http://docs.rackspace.com/orchestration/api/v1/orchestration-devguide/content/POST_stack_adopt__v1__tenant_id__stacks_Stack_Operations.html) and tested against devstack and Rackspace public endpoints. I know the documentation may not reflect reality, but, if not, we need to know that.

pratikmallya:
@jrperritt good point. I will add comments so its clearer why I made the changes.

openstack/orchestration/v1/stackresources/results.go
```diff
@@ -75,18 +86,6 @@ func (r ResourcePage) IsEmpty() (bool, error) {
 	return len(resources) == 0, nil
 }
 
-// LastMarker returns the last container name in a ListResult.
```
pratikmallya: This change should have been in https://github.com/rackspace/gophercloud/pull/466. Apologies

openstack/orchestration/v1/stackresources/results.go
```diff
@@ -192,21 +206,40 @@ func (r ResourceTypePage) IsEmpty() (bool, error) {
 	return len(rts) == 0, nil
 }
 
+// resourceTypes represents the type that holds the result of ExtractResourceTypes.
+// We define methods on this type to sort it before output
+type resourceTypes []string
```
pratikmallya: as the comment says, I added this interface simply to enable sorting. This is because the output from `heatclient` is sorted...and also because it looks _much_ nicer. However, we can possibly do the sorting in the code for the commands itself. I'm ok with either way.

openstack/orchestration/v1/stackresources/results.go
```diff
@@ -230,14 +263,6 @@ func (r SchemaResult) Extract() (*TypeSchema, error) {
 	return &res, nil
 }
 
-// TypeTemplate represents a stack resource template.
-type TypeTemplate struct {
```
pratikmallya: I removed this because the output of http://api.rackspace.com/api-ref-orchestration.html#resource_type_template is a heat template, a string. It doesn't make much sense to parse a heat template, as its considered to be an atomic unit of data for orchestration.

openstack/orchestration/v1/stacks/requests.go
```diff
@@ -133,20 +139,6 @@ type AdoptOpts struct {
 	Name string
 	// (REQUIRED) The timeout for stack creation in minutes.
 	Timeout int
-	// (OPTIONAL; REQUIRED IF Template IS EMPTY) The URL of the template to instantiate.
```
pratikmallya: `stack-adopt` has been changed so that it does not accept `templateURL` or `template`. This change was made long ago but the documentation has not been updated. I will file a bug. The heatclient does not accept either `templateURL` or `template`. This commit: https://github.com/openstack/python-heatclient/commit/a34f105ce1096dbdd35568f2289119fefd686701 shows when the option was removed from heatclient

openstack/orchestration/v1/stacks/requests.go
```diff
@@ -197,12 +181,12 @@ func (opts AdoptOpts) ToStackAdoptMap() (map[string]interface{}, error) {
...
 	}
 
-	if opts.Timeout == 0 {
-		return nil, errors.New("Required field 'Timeout' not provided.")
+	if opts.Timeout != 0 {
+		s["timeout"] = opts.Timeout
 	}
 	s["timeout_mins"] = opts.Timeout
 
-	return map[string]interface{}{"stack": s}, nil
```
pratikmallya: Maybe they changed the request structure sometimes soon? Making a request to the api will not work unless this change is made

openstack/orchestration/v1/stacks/results.go
```diff
@@ -200,21 +202,19 @@ type DeleteResult struct {
...
-	Description         string                   `mapstructure:"description"`
-	DisableRollback     bool                     `mapstructure:"disable_rollback"`
-	ID                  string                   `mapstructure:"id"`
-	Links               []gophercloud.Link       `mapstructure:"links"`
-	Name                string                   `mapstructure:"stack_name"`
-	NotificationTopics  []interface{}            `mapstructure:"notification_topics"`
-	Parameters          map[string]string        `mapstructure:"parameters"`
-	Resources           []map[string]interface{} `mapstructure:"resources"`
-	Status              string                   `mapstructure:"stack_status"`
-	StatusReason        string                   `mapstructure:"stack_status_reason"`
```
pratikmallya: `stack preview` gives the user an idea of what resource will be deployed. Since there is no stack (or resources) actually created, they have neither a `Status` nor `StatusReason` http://api.rackspace.com/api-ref-orchestration.html#stack_preview

openstack/orchestration/v1/stacktemplates/requests.go
```diff
@@ -23,14 +23,14 @@ type ValidateOptsBuilder interface {
 
 // ValidateOpts specifies the template validation parameters.
 type ValidateOpts struct {
-	Template    map[string]interface{}
+	Template    string
```
pratikmallya: The template should be a string, since the api expects a string

openstack/orchestration/v1/stacktemplates/results.go
```diff
@@ -1,42 +1,33 @@
 package stacktemplates
 
 import (
+	"encoding/json"
 	"github.com/mitchellh/mapstructure"
 	"github.com/rackspace/gophercloud"
 )
 
-// Template represents a stack template.
```
pratikmallya: Same as my previous comment. The template must not be parsed.

pratikmallya:
@jrperritt I added comments to the code where I changed the interface significantly. Let me know if you need more context.

openstack/orchestration/v1/stacks/requests.go
pratikmallya: Actually, it looks like the api still accepts `template` and `templateURL`, but gives preference to `adopt-stack-data`. I'll confirm this tomorrow. If that's the case, we can leave this interface as-is, and simply not provide options for `template` or `templateURL` in the commands in rack.

pratikmallya:
@jrperritt can you please take a look?

openstack/orchestration/v1/stacks/requests.go
pratikmallya: I decided not to remove the fields in `AdoptOpts`. It should be changed only if/when the documented api is changed

pratikmallya:
Closing, since commit is included in https://github.com/rackspace/gophercloud/pull/486

PR #484, opened 2015-09-14
[rfr]Make client return error on JSON decoding error

pratikmallya:
If the response is not valid JSON, the request simply returns an
empty body. If the user is expecting the result to be JSON and its
not, we should signal that an error has occured.

### DIFF
--- a/openstack/compute/v2/servers/requests.go
+++ b/openstack/compute/v2/servers/requests.go
@@ -754,9 +754,7 @@ func Metadatum(client *gophercloud.ServiceClient, id, key string) GetMetadatumRe
 // DeleteMetadatum will delete the key-value pair with the given key for the given server ID.
 func DeleteMetadatum(client *gophercloud.ServiceClient, id, key string) DeleteMetadatumResult {
 	var res DeleteMetadatumResult
-	_, res.Err = client.Delete(metadatumURL(client, id, key), &gophercloud.RequestOpts{
-		JSONResponse: &res.Body,
-	})
+	_, res.Err = client.Delete(metadatumURL(client, id, key), nil)
 	return res
 }
 
--- a/rackspace/networking/v2/networks/delegate_test.go
+++ b/rackspace/networking/v2/networks/delegate_test.go
@@ -203,8 +203,17 @@ func TestCreateWithOptionalFields(t *testing.T) {
 	}
 }
 		`)
-
 		w.WriteHeader(http.StatusCreated)
+		fmt.Fprintf(w, `
+{
+	"network": {
+			"name": "sample_network",
+			"admin_state_up": true,
+			"shared": true,
+			"tenant_id": "12345"
+	}
+}
+		`)
 	})
 
 	iTrue := true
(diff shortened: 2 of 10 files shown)


pratikmallya:
@jrperritt I have a suspicion that since this patch makes errors bubble up, its breaking other tests that were using invalid JSON's in their requests.

jrperritt:
Hmm, possibly. I'll be interested to see the changes since, as far as I know, we don't explicitly pass JSON objects, but instead create `map`s, which should conform to the JSON spec when marshalled.

rackspace/networking/v2/networks/delegate_test.go
```diff
@@ -203,8 +203,17 @@ func TestCreateWithOptionalFields(t *testing.T) {
 	}
 }
 		`)
-
 		w.WriteHeader(http.StatusCreated)
+		fmt.Fprintf(w, `
```
pratikmallya: @jrperritt so, this is an instance of what I was talking about. The test failed here because the patch returns an error if the response is empty (`unexpected error "EOF"`) but this change fixes the test.

jrperritt: :+1:

pratikmallya:
@jrperritt updated unit tests so they pass. There was also a minor issue with deleting server metadata that I fixed.

openstack/compute/v2/servers/fixtures.go
```diff
@@ -1,7 +1,7 @@
-// +build fixtures
```
pratikmallya: `go test` has a bug where unless the `package` statement is the top, it won't import variables into other files in the same package.

openstack/compute/v2/servers/requests.go
```diff
@@ -754,9 +754,7 @@ func Metadatum(client *gophercloud.ServiceClient, id, key string) GetMetadatumRe
 // DeleteMetadatum will delete the key-value pair with the given key for the given server ID.
 func DeleteMetadatum(client *gophercloud.ServiceClient, id, key string) DeleteMetadatumResult {
 	var res DeleteMetadatumResult
-	_, res.Err = client.Delete(metadatumURL(client, id, key), &gophercloud.RequestOpts{
-		JSONResponse: &res.Body,
-	})
+	_, res.Err = client.Delete(metadatumURL(client, id, key), nil)
```
pratikmallya: `delete`  has no response body http://api.rackspace.com/api-ref.html#deleteMetadataItem

openstack/compute/v2/servers/fixtures.go
jrperritt: The build tag has to be at the top. Try using 
`go test -tags fixtures github.com/rackspace/gophercloud/openstack/compute/v2/servers`
to run those tests.

jrperritt: Alternatively, to run all the Gophercloud unit tests, there is a script in the root directory.
Run `script/unittest` from the command-line

pratikmallya: Thanks, I undid that change.

jrperritt:
Great fixes

+2

PR #486, opened 2015-09-14
[rfr]Enable template environment parsing

pratikmallya:


pratikmallya:
Added changes to acceptance tests as well

pratikmallya:
Bump.Any updates on this?

jrperritt:
Reviewing now

jrperritt:
Let's amend `.travis.yml` to remove `1.1` and add `1.5`.

openstack/orchestration/v1/stacks/utils.go
```diff
@@ -0,0 +1,145 @@
...
+
+	"github.com/rackspace/gophercloud"
+	"gopkg.in/yaml.v2"
+)
+
+type Client interface {
+	Get(string) (*http.Response, error)
+}
+
+type TE struct {
```
jrperritt: What is a `TE`?

pratikmallya: @jrperritt its meant to be the base structure for both a template and an environment.

openstack/orchestration/v1/stacks/template.go
```diff
@@ -0,0 +1,113 @@
...
+	TE
+}
+
+var TemplateFormatVersions = map[string]bool{
+	"HeatTemplateFormatVersion": true,
+	"heat_template_version":     true,
+	"AWSTemplateFormatVersion":  true,
+}
+
+func (t *Template) Validate() error {
```
jrperritt: This is really nice.

openstack/orchestration/v1/stacks/template.go
```diff
@@ -0,0 +1,113 @@
...
+	}
+	for key, _ := range t.Parsed {
+		if _, ok := TemplateFormatVersions[key]; ok {
+			return nil
+		}
+	}
+	return errors.New(fmt.Sprintf("Template format version not found."))
+}
+
+func GetFileContents(t *Template, te interface{}, ignoreIf igFunc, recurse bool) error {
```
jrperritt: Any reason to not have this as a method on a `*Template` like `Validate`?

pratikmallya: @jrperritt not really. Is it a good idiom in go to have most functions as methods?

jrperritt: The logic in this function is hard to follow. As a rule, all exported functions should have comments about what they do. Complex functions like this should have detailed comments about how it works.

openstack/orchestration/v1/stacks/utils.go
```diff
@@ -0,0 +1,145 @@
...
+	Files map[string]string
+	// fileMaps is a map used internally when determining Files.
+	fileMaps map[string]string
+	// baseURL represents the location of the template or environment file.
+	baseURL string
+	// client is an interface which allows TE to fetch contents from URLS
+	client Client
+}
+
+func (t *TE) Fetch() error {
```
jrperritt: Detailed comment for this method. Why would I use it? How would I use it?

I know hardly any of the other functions/methods in Gophercloud have detailed comments like I'm requesting, but that's because they're small and it's obvious what they do. Heat by design is probably the most complex service and, as such, deserves more documentation.

openstack/orchestration/v1/stacks/utils.go
```diff
@@ -0,0 +1,145 @@
...
+	"net/http"
+	"path/filepath"
+	"reflect"
+	"strings"
+
+	"github.com/rackspace/gophercloud"
+	"gopkg.in/yaml.v2"
+)
+
+type Client interface {
```
jrperritt: Detailed comment for this interface. Why would I use it? How would I use it?

openstack/orchestration/v1/stacks/template.go
jrperritt: In re ^^: Not necessarily, but if you're passing the object (that is defined in the same package) as a parameter anyway, I think it makes sense.

jrperritt:
@pratikmallya This looks _really_ good. It's almost exclusively my ignorance of Heat that is causing the delays. I'll make another pass to review later today.

pratikmallya:
@jrperritt no worries. I will add more documentation to make it clearer what the functions are supposed to do.

openstack/orchestration/v1/stacks/template.go
pratikmallya: Done.

openstack/orchestration/v1/stacks/utils.go
pratikmallya: Done.

openstack/orchestration/v1/stacks/utils.go
pratikmallya: Done.

pratikmallya:
Documentation that might help with review:
https://github.com/openstack/heat/blob/master/doc/source/template_guide/environment.rst : describes what an environment file in heat looks like.
https://github.com/openstack/python-heatclient/blob/master/heatclient/common/template_utils.py : I used the parser here as a model for implementing the parser in this patch. I've tried to keep the behavior of both exactly the same.

openstack/orchestration/v1/stacks/environment.go
```diff
@@ -0,0 +1,137 @@
...
+			return errors.New(fmt.Sprintf("Environment has wrong section: %s", key))
+		}
+	}
+	return nil
+}
+
+// Parse environment file to resolve the URL's of the resources. This is done by
+// reading from the `Resource Registry` section, which is why the function is
+// named GetRRFileContents.
+func (e *Environment) GetRRFileContents(ignoreIf igFunc) error {
```
jrperritt: Will users be calling this method? If not, it should be lowercase (`getRRFileContents`). If so, `igFunc` should be exported (`IgFunc`), too. Otherwise, I don't think users will be able to create that type.

pratikmallya: @jrperritt good point. I've made the parsing methods local

jrperritt:
@pratikmallya This looks good to me. The next thing is to run `go vet` and address any warnings it gives (only the ones that pertain to this PR).

pratikmallya:
@jrperritt  I don't see any output from `go vet`

jrperritt:
Sorry. `go vet ./...`

pratikmallya:
`(py275)MNK0AKG3QN:gophercloud prat8228$ go vet ./...
(py275)MNK0AKG3QN:gophercloud prat8228$`

pratikmallya:
@jrperritt again, no output :)

jrperritt:
@pratikmallya thank you for the fixes

+2

PR #488, opened 2015-09-16
Allowed address pairs support for Neutron Port

trumant:
This is my first contribution to the project so very open to and looking forward to any constructive feedback

### DIFF
--- a/openstack/networking/v2/ports/requests.go
+++ b/openstack/networking/v2/ports/requests.go
@@ -168,12 +168,13 @@ type UpdateOptsBuilder interface {
 
 // UpdateOpts represents the attributes used when updating an existing port.
 type UpdateOpts struct {
-	Name           string
-	AdminStateUp   *bool
-	FixedIPs       interface{}
-	DeviceID       string
-	DeviceOwner    string
-	SecurityGroups []string
+	Name                string
+	AdminStateUp        *bool
+	FixedIPs            interface{}
+	DeviceID            string
+	DeviceOwner         string
+	SecurityGroups      []string
+	AllowedAddressPairs interface{}
 }
 
 // ToPortUpdateMap casts an UpdateOpts struct to a map.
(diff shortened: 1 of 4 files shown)


openstack/networking/v2/ports/requests.go
```diff
@@ -168,12 +168,13 @@ type UpdateOptsBuilder interface {
...
-	DeviceID       string
-	DeviceOwner    string
-	SecurityGroups []string
+	Name                string
+	AdminStateUp        *bool
+	FixedIPs            interface{}
+	DeviceID            string
+	DeviceOwner         string
+	SecurityGroups      []string
+	AllowedAddressPairs interface{}
```
pratikmallya: since the structure of `AllowedAddressPairs` is known, it would be better to specify this as type `[]map[string]string`

trumant: I was just following the pattern currently in place for FixedIPs

trumant:
Assuming this looks good to merge, I'll add support for this option on Port creation as well. Just wanted to get a complete, working version of update through review before going too far down the rabbit hole on this.

nmische:
Any update on this? We would really like to see allowed address pairs added to gophercloud such that we can provision via Terraform.

trumant:
@jrperritt - Any pointers you could give me about getting this to a mergeable point would be appreciated.

openstack/networking/v2/ports/requests.go
jrperritt: I'm not sure why we used an `interface{}` for the `FixedIPs`. We may not have found documentation for them when that was written (i.e. is it an array of objects, an array of strings, something else, etc). Regardless, let's use `[]AddressPair` if we know that's the format. Same thing on line 107.

jrperritt:
It looks good aside from the 2 minor type changes. I'm not concerned with the unit tests failing on the Go tip branch.

trumant:
Thanks for the feedback @jrperritt I'll work to address it.

roobert:
Hi, Like @nmische, I came here looking for this feature since it's required in order to update the Terraform OpenStack provider. Since this ticket hasn't been updated in 20 days I was wondering if there is anything I could do to help get this merged?

Cheers,

trumant:
@nmische and @roobert and @jrperritt - Finally got around to making this change. Appreciate the reminder and feedback.

roobert:
Thanks @trumant! Hopefully @jrperritt will be able to merge this now.

jrperritt:
@trumant looks great. +2

PR #489, opened 2015-09-17
[rfr] Add volumes part of blockstorage v2 API

feiskyer:
As discussed in #485 , split blockstorage v2 API into a few parts. This is volumes API.


openstack/blockstorage/v2/volumes/requests.go
```diff
@@ -0,0 +1,418 @@
...
+type CreateOptsBuilder interface {
+	ToVolumeCreateMap() (map[string]interface{}, error)
+}
+
+// CreateOpts contains options for creating a Volume. This object is passed to
+// the volumes.Create function. For more information about these parameters,
+// see the Volume object.
+type CreateOpts struct {
+	// The availability zone [OPTIONAL]
+	Availability string
```
jtopjian: Thoughts about changing this to  `AvailabilityZone`? I know it was `Availability` in v1, but maybe now is a good time to make the name more intuitive.

openstack/blockstorage/v2/volumes/requests.go
```diff
@@ -0,0 +1,418 @@
...
+	// he volume description [OPTIONAL]
```
jtopjian: `s/he/The/`

openstack/blockstorage/v2/volumes/requests.go
```diff
@@ -0,0 +1,418 @@
...
+// CreateOptsBuilder allows extensions to add additional parameters to the
+// Create request.
+type CreateOptsBuilder interface {
+	ToVolumeCreateMap() (map[string]interface{}, error)
+}
+
+// CreateOpts contains options for creating a Volume. This object is passed to
+// the volumes.Create function. For more information about these parameters,
+// see the Volume object.
+type CreateOpts struct {
```
jtopjian: Even though these parameters are not explicitly mentioned in the API docs table, they are mentioned in the example JSON request. They are also referenced in the actual API code under the `create` function:

* SourceReplica
* ConsistencyGroupID

openstack/blockstorage/v2/volumes/requests.go
```diff
@@ -0,0 +1,418 @@
...
+	SnapshotID string
+	// the ID of the existing volume [OPTIONAL]
+	SourceVolID string
+	// The ID of the image from which you want to create the volume.
+	// Required to create a bootable volume.
+	ImageID string
+	// The associated volume type [OPTIONAL]
+	VolumeType string
+	// The unique identifier of the tenant or account [OPTIONAL]
+	TenantID string
```
jtopjian: I may be mistaken, but I don't believe TenantID is a valid create parameter. The Tenant ID is used in the URL to make the request, but AFAIK, it's not a parameter that can be set during creation.

openstack/blockstorage/v2/volumes/requests.go
```diff
@@ -0,0 +1,418 @@
...
+// the Volume object.
+type UpdateOpts struct {
+	// OPTIONAL
+	Name string
+	// OPTIONAL
+	Description string
+	// OPTIONAL
+	Metadata map[string]string
+	// OPTIONAL
+	TenantID string
```
jtopjian: As with `Create`, unless I'm mistaken, TenantID can't be set.

openstack/blockstorage/v2/volumes/requests.go
```diff
@@ -0,0 +1,418 @@
...
+	case 0:
+		return "", fmt.Errorf("Unable to find volume: %s", name)
+	case 1:
+		return volumeID, nil
+	default:
+		return "", fmt.Errorf("Found %d volumes matching %s", volumeCount, name)
+	}
+}
+
+type AttachOptsBuilder interface {
```
jtopjian: I'm very sorry to keep harping on this, but I really do think that everything from here to the end of the file needs to be moved to a `volumeactions` package. If I'm incorrect, I'm doubly sorry, but I can't find anything that shows that `os-volume-actions` is a core part of `volumes` and not an extension.

openstack/blockstorage/v2/volumes/requests_test.go
```diff
@@ -0,0 +1,197 @@
...
+		if err != nil {
+			t.Errorf("Failed to extract volumes: %v", err)
+			return false, err
+		}
+
+		expected := []Volume{
+			{
+				ID:                        "289da7f8-6440-407c-9fb4-7db01ec49164",
+				Name:                      "vol-001",
+				Attachments:               []map[string]interface{}{},
```
jtopjian: Optional, but I think it would be beneficial to have some real attachment data here so tests can verify if it is correctly being parsed.

openstack/blockstorage/v2/volumes/requests_test.go
```diff
@@ -0,0 +1,197 @@
...
+				AvailabilityZone:          "nova",
+				Bootable:                  "false",
+				ConsistencyGroupID:        "",
+				CreatedAt:                 "2015-09-17T03:35:03.000000",
+				Description:               "",
+				Encrypted:                 false,
+				Metadata:                  map[string]string{},
```
jtopjian: Similar to Attachments, I think having one piece of metadata would help ensure parsing is correct.

openstack/blockstorage/v2/volumes/results.go
```diff
@@ -0,0 +1,167 @@
...
+	// Current status of the volume.
+	Status string `mapstructure:"status"`
+
+	// Human-readable display name for the volume.
+	Name string `mapstructure:"name"`
+
+	// Instances onto which the volume is attached.
+	Attachments []map[string]interface{} `mapstructure:"attachments"`
+
+	// This parameter is no longer used.
```
jtopjian: I think this is incorrect?

openstack/blockstorage/v2/volumes/requests.go
feiskyer: OK

openstack/blockstorage/v2/volumes/requests_test.go
feiskyer: Thanks, I'll add that

openstack/blockstorage/v2/volumes/requests_test.go
feiskyer: Thanks, I'll add that

openstack/blockstorage/v2/volumes/results.go
feiskyer: Thanks for figured out.

openstack/blockstorage/v2/volumes/requests.go
feiskyer: @jtopjian Yes, you are right, although it is listed at documentation.

feiskyer:
@jtopjian Committed as your suggestion.

feiskyer:
@jrperritt Please review

jrperritt:
@feiskyer Before review, please address failing tests.

feiskyer:
@jrperritt It failed because acceptance tests has no environment variables: "Environment variable OS_AUTH_URL needs to be set."

jtopjian:
I think it may be failing for another reason. I recently pushed some code that included acceptance tests and Travis did not fail for me.

Comparing the v1 and v2 versions of the acceptance tests, the following line is missing from the top of the v2 test:

```
// +build acceptance blockstorage
```

I'm not sure if that causes tests to be run conditionally depending on the environment.

Another reason why it's useful to include the diff from the v1 code.

jrperritt:
@feiskyer Exactly what @jtopjian said. In addition, after a PR review has started, unless specifically asked, **never** squash commits. It makes it much more difficult for the reviewer to see what's changed.

feiskyer:
@jtopjian @jrperritt Thanks for the tip, all checks have passed.

feiskyer:
CC @jrperritt

feiskyer:
CC @jrperritt again

jtopjian:
@feiskyer Is it possible to create commits that include the diffs from v1? That might make reviewing this easier. The way it's committed now, it feels like all of the code is brand new, though I imagine only a small percentage is actually new.

feiskyer:
@jtopjian Not easy to include diffs from v1. v2 is in different package from v1.

jtopjian:
@feiskyer See https://github.com/jtopjian/gophercloud/tree/openstack-blockstorage-v2. This was made by first copying v1 to v2 and then making the appropriate changes.

feiskyer:
@jtopjian @jrperritt Updated commits.

feiskyer:
CC @jrperritt  ?

clintonskitson:
@feiskyer I was testing out your patch with the force detach feature for cinder v2.  I ran into a couple of problems.  First, the attach method described based on the docs seems to actually be forcedetach.  Referencing the API docs below.  I believe you will find the detach in the compute extensions.
http://developer.openstack.org/api-ref-blockstorage-v2.html

Below is my working patch to make force_detach work.

https://github.com/clintonskitson/gophercloud/blob/2c94a0fa5694051e79a7b817aa9ceeeb43d960c0/openstack/blockstorage/v2/extensions/volumeactions/requests.go#L57-L69

The second piece that I ran into was the results from this call.  I'm not sure if it's correct or not, but the return was "null" in the body which was being interpreted as an error.

https://github.com/clintonskitson/gophercloud/blob/feature/add_force_detach/provider_client.go#L231-L234

feiskyer:
@clintonskitson `os-detach` is not same as `os-force_detach`:

[os-force_detach](https://github.com/openstack/cinder/blob/master/cinder/api/contrib/admin_actions.py#L180-L216) is a part of admin actions:

```python
    @wsgi.action('os-force_detach')
    def _force_detach(self, req, id, body):
        """Roll back a bad detach after the volume been disconnected."""
```

while [os-detach](https://github.com/openstack/cinder/blob/master/cinder/api/contrib/volume_actions.py#L134-L160) is a part of volume actions:

```python
    @wsgi.action('os-detach')
    def _detach(self, req, id, body):
        """Clear attachment metadata."""
        context = req.environ['cinder.context']
```

CC  @jrperritt . I think it's been a long for this PR to review.

clintonskitson:
Interesting @feiskyer.  Not seeing that called out in the docs.  The force_detach would be additive then as a useful feature.  The second portion of what I called out needs to be addressed as well I believe.

feiskyer:
@jrperritt Could you pls review this?

openstack/blockstorage/v2/extensions/volumeactions/requests.go
```diff
@@ -0,0 +1,176 @@
+package volumeactions
+
+import (
+	"github.com/rackspace/gophercloud"
+)
+
+type AttachOptsBuilder interface {
```
jamiehannaford: Each exported type and function in this file needs an accompanying docblock, like [this](https://github.com/rackspace/gophercloud/blob/master/openstack/blockstorage/v1/volumes/requests.go#L10-L11).

openstack/blockstorage/v2/extensions/volumeactions/requests.go
```diff
@@ -0,0 +1,176 @@
...
+
+type AttachOpts struct {
+	// The mountpoint of this volume
+	MountPoint string
+	// The nova instance ID, can't set simultaneously with HostName
+	InstanceUUID string
+	// The hostname of baremetal host, can't set simultaneously with InstanceUUID
+	HostName string
+	// Mount mode of this volume
+	Mode string
```
jamiehannaford: will this only ever be `ro` or `rw`, or do we expect custom values? if the former, perhaps an enum is better:

```go
// Mode describes the attachment mode for volumes.
type Mode string

// These constants determine how a volume is attached
const (
	ReadOnly Mode = "ro"
	ReadWrite Mode = "rw"
)

type AttachOpts struct {
	...
	Mode Mode
	...
}
```

openstack/blockstorage/v2/extensions/volumeactions/requests.go
```diff
@@ -0,0 +1,176 @@
...
+	var res AttachResult
+
+	reqBody, err := opts.ToVolumeAttachMap()
+	if err != nil {
+		res.Err = err
+		return res
+	}
+
+	_, res.Err = client.Post(attachURL(client, id), reqBody, &res.Body, &gophercloud.RequestOpts{
+		OkCodes: []int{200, 201, 202},
```
jamiehannaford: I think the API [returns 202](https://github.com/openstack/cinder/blob/master/cinder/api/contrib/volume_actions.py#L93)

openstack/blockstorage/v2/extensions/volumeactions/requests.go
```diff
@@ -0,0 +1,176 @@
...
+}
+
+func Detach(client *gophercloud.ServiceClient, id string) DetachResult {
+	var res DetachResult
+
+	v := make(map[string]interface{})
+	reqBody := map[string]interface{}{"os-detach": v}
+
+	_, res.Err = client.Post(detachURL(client, id), reqBody, &res.Body, &gophercloud.RequestOpts{
+		OkCodes: []int{200, 201, 202},
```
jamiehannaford: [202](https://github.com/openstack/cinder/blob/master/cinder/api/contrib/volume_actions.py#L121)

openstack/blockstorage/v2/extensions/volumeactions/requests.go
```diff
@@ -0,0 +1,176 @@
...
+
+type ConnectorOptsBuilder interface {
+	ToConnectorMap() (map[string]interface{}, error)
+}
+
+type ConnectorOpts struct {
+	IP        string
+	Host      string
+	Initiator string
+	Wwpns     string
```
jamiehannaford: Are you sure this is a string? I've found [test fixtures](https://github.com/openstack/cinder/blob/master/cinder/tests/unit/test_dothill.py#L132) which structures it as `[]string`. I can't find any authoritative docs

jamiehannaford:
@feiskyer In [another PR](#561), @j-griffith mentioned that volume actions are version independent, but I can't find any docs which verify that. So if that's the case, we'll need to move that package out of the v2 dir.

j-griffith:
Hmm, I hadn't seen this PR; looks like I needlessly duplicated some effort.

So yeah, we could/should probably move the extensions in to their own module, currently in Cinder we don't associate them with Versions.  You can look here and get a view of how these are laid out:
https://github.com/openstack/cinder/tree/master/cinder/api/contrib

Note that the contrib directory is imported by all of the versions and doesn't have any overrides on the actions.  

Bummer is that as of this week here at the OS design summit it sounds like those are going to be deprecated and placed under versioned API modules, but we can deal with that much later down the road.

jamiehannaford:
If the long-term plan is to have versioned extensions, I think it's probably better to keep the above volume actions under a `v2` directory, and implement some kind of common logic whenever v3+ comes into existence. For me that's preferable to storing stuff under a location that will soon be deprecated and will be potentially confusing to future users.

j-griffith:
@jamiehannaford Up to ya'll, but I think that's the wrong way to go.  The versioning move will put them ALL under V3 so this would all just be wrong both before and after the micro version changes merge.  Not to mention the deprecation period is 2 releases, so we're a year out anyway.

Anyway, I'm not overly partial one way or the other.  Just suggesting correct might be better, and also don't want to lock out anybody that may choose to still use V1 of the Cinder API.  Up until this last release Nova and some other's still use V1.

As the micro version changes form up and start to take shape in Cinder I'm happy to update things here if that helps at all.

openstack/blockstorage/v2/extensions/volumeactions/requests.go
feiskyer: Good point. thanks.

openstack/blockstorage/v2/extensions/volumeactions/requests.go
feiskyer: Good point. thanks.

openstack/blockstorage/v2/extensions/volumeactions/requests.go
feiskyer: Yes, it's my mistake.

feiskyer:
@jamiehannaford Updated docs and bug fixes.

For the version problem, I think it's ok to keep those operations in v2 pkg, and refactor or rewrite after v3 is released.

acceptance/openstack/blockstorage/v2/extensions/volumeactions_test.go
```diff
@@ -0,0 +1,59 @@
...
+		err = volumes.WaitForStatus(client, cv.ID, "available", 60)
+		th.AssertNoErr(t, err)
+		err = volumes.Delete(client, cv.ID).ExtractErr()
+		th.AssertNoErr(t, err)
+	}()
+
+	err = volumes.WaitForStatus(client, cv.ID, "available", 60)
+	th.AssertNoErr(t, err)
+
+	_, err = volumeactions.Attach(client, cv.ID, &volumeactions.AttachOpts{
```
jamiehannaford: I think you need a step to work out the instance ID, something like:

```go
instanceID := os.Getenv("OS_INSTANCE_ID")
if !instanceID {
	// throw error or create an instance
}
```

acceptance/openstack/blockstorage/v2/extensions/volumeactions_test.go
```diff
@@ -0,0 +1,59 @@
...
+		MountPoint:   "/mnt",
+		Mode:         "rw",
+		InstanceUUID: "50902f4f-a974-46a0-85e9-7efc5e22dfdd",
```
jamiehannaford: need to replace this hard-coded UUID with `instanceID`

acceptance/openstack/blockstorage/v2/extensions/volumeactions_test.go
```diff
@@ -0,0 +1,59 @@
...
+	}).Extract()
+	th.AssertNoErr(t, err)
+
+	err = volumes.WaitForStatus(client, cv.ID, "in-use", 60)
+	th.AssertNoErr(t, err)
+
+	_, err = volumeactions.Detach(client, cv.ID).Extract()
+	th.AssertNoErr(t, err)
+}
```
jamiehannaford: Is it possible to add acceptance tests for reserve, unreserve, init conn and terminate conn? Or is the set-up too fiddly?

jamiehannaford:
@feiskyer I've left a last round of feedback, sorry for the delay. If you could reply to those comments above, I can merge.

acceptance/openstack/blockstorage/v2/extensions/volumeactions_test.go
feiskyer: Thanks, updated.

acceptance/openstack/blockstorage/v2/extensions/volumeactions_test.go
feiskyer: Thanks, updated.

acceptance/openstack/blockstorage/v2/extensions/volumeactions_test.go
feiskyer: Acceptance test for volume reserve/unreserve added. But volumeactions.InitializeConnection and volumeactions.TerminateConnection is harder to test, because the response data is depending on cinder volume driver.

feiskyer:
@jamiehannaford Thanks for review. Updated as comments. Please review it again.

acceptance/openstack/blockstorage/v2/extensions/volumeactions_test.go
jtopjian: Would it be safe to assume the standard LVM driver for Cinder? This is what is used in a default devstack installation.

feiskyer: It depends. I always run with ceph.

jamiehannaford:
@feiskyer I ran into some `EOF` errors when running the acceptance tests because empty HTTP responses were trying to be JSON decoded. I [submitted a PR](https://github.com/feiskyer/gophercloud/pull/1) to your branch which fixes this. I also added the acceptance tests for init/terminate connections.

feiskyer:
@jamiehannaford Thanks a lot. Got merged.

PR #490, opened 2015-09-17
[rfr]Add test for detecting misformatted code

pratikmallya:
Also fixes formatting in code that was not formatted correctly

### DIFF
--- a/acceptance/openstack/pkg.go
+++ b/acceptance/openstack/pkg.go
@@ -1,4 +1,3 @@
 // +build acceptance
 
 package openstack
-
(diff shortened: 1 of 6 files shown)


jrperritt:
This is a good idea. Is this just a unit test that will cause Travis to fail if there are any formatting issues?

pratikmallya:
@jrperritt exactly

pratikmallya:
@jrperritt perhaps a better way to do this would be to add this to the travis test itself? I'm kinda leaning towards that instead of having a unit test like this one

jrperritt:
That's what I'm thinking, too

jrperritt:
Closing in favor of checking during build.

PR #491, opened 2015-09-17
Method for finding a node by IP and Port

rgbkrk:
As suggested by @jrperritt, I've extracted FindNodeByIPAndPort out of [peekaboo](https://github.com/rgbkrk/peekaboo).

/cc @jamiehannaford @smashwilson as well for feedback on style nits, naming, and/or your favorite pony color.

### DIFF
--- a/rackspace/lb/v1/nodes/requests.go
+++ b/rackspace/lb/v1/nodes/requests.go
@@ -279,6 +279,9 @@ func GetByIPPort(
 	})
 
 	// TODO: When found is nil, return an error
+	if found == nil {
+		return nil, errors.New("Unable to get node by IP and Port")
+	}
 
 	return found, nil
 }
--- a/rackspace/lb/v1/nodes/requests.go
+++ b/rackspace/lb/v1/nodes/requests.go
@@ -278,7 +278,6 @@ func GetByIPPort(
 		return true, nil
 	})
 
-	// TODO: When found is nil, return an error
 	if found == nil {
 		return nil, errors.New("Unable to get node by IP and Port")
 	}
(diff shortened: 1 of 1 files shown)


rackspace/lb/v1/nodes/requests.go
```diff
@@ -249,3 +249,36 @@ func ListEvents(client *gophercloud.ServiceClient, loadBalancerID int, opts List
 		return NodeEventPage{pagination.SinglePageBase(r)}
 	})
 }
+
+// FindNodeByIPPort locates a load balancer node by IP and port.
+func FindNodeByIPPort(
```
jrperritt: Since it's in the `nodes` package, we can drop the "Node" part of the name. Also, for convention, let's go with "Get" instead of "Find".
s/FindNodeByIPPort/GetByIPPort

rackspace/lb/v1/nodes/requests.go
```diff
@@ -249,3 +249,36 @@ func ListEvents(client *gophercloud.ServiceClient, loadBalancerID int, opts List
...
+				found = &trialNode
+				return false, nil
+			}
+
+		}
+
+		return true, nil
+	})
+
+	// TODO: When found is nil, return an error
```
jrperritt: Good idea

rgbkrk:
Addressed initial comments.

jrperritt:
Are you still working on this? It looks good to me.

rgbkrk:
I certainly haven't touched it in a while. If it fits it ships.

jrperritt:
+2

PR #492, opened 2015-09-18
[rfr] Updating Compute documentation

jtopjian:
Scratching an itch from #472.

Most examples were just pulled from the acceptance tests. I'm running blind with the syntax formatting since Github doesn't render the `highlight` style formatting.


jrperritt:
Thank you for updating the docs @jtopjian

PR #494, opened 2015-09-22
[rfr]Support Extra routes extension in router api

timbyr:
http://developer.openstack.org/api-ref-networking-v2-ext.html#extraroute-ext

### DIFF
--- a/openstack/networking/v2/extensions/layer3/routers/requests.go
+++ b/openstack/networking/v2/extensions/layer3/routers/requests.go
@@ -97,6 +97,7 @@ type UpdateOpts struct {
 	Name         string
 	AdminStateUp *bool
 	GatewayInfo  *GatewayInfo
+	Routes       []Route
 }
 
 // Update allows routers to be updated. You can update the name, administrative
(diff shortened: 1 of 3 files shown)


openstack/networking/v2/extensions/layer3/routers/requests.go
```diff
@@ -97,6 +97,7 @@ type UpdateOpts struct {
 	Name         string
 	AdminStateUp *bool
 	GatewayInfo  *GatewayInfo
+	Routes       *[]Route
```
jrperritt: Slices are pointers underneath, so we can just have `[]Route` here.

jrperritt:
@timbyr Just one small change, then we can merge.

hingstarne:
Hi, 
are there any news on that. We could really have a big advantage of having these feature.
Cheers

Fodoj:
+1 - would be a-mazing to see this PR merged

jrperritt:
Sorry for the delay; this looks good. +2

PR #496, opened 2015-09-23
Add PrettyPrintJSON method for gophercloud.Link structure

pratikmallya:
This helper method is useful when the contents need to be
displayed.

### DIFF
--- a/results.go
+++ b/results.go
@@ -129,6 +129,16 @@ type Link struct {
 	Rel  string `mapstructure:"rel"`
 }
 
+// PrettyPrintJSON creates a string containing the links as
+// pretty-printed JSON. It is useful for displaying the results.
+func (link Link) PrettyPrintJSON() string {
+	pretty, err := json.MarshalIndent(link, "", "  ")
+	if err != nil {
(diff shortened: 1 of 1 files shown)


results.go
```diff
@@ -129,6 +129,16 @@ type Link struct {
 	Rel  string `mapstructure:"rel"`
 }
 
+// PrettyPrintJSON creates a string containing the links as
+// pretty-printed JSON. It is useful for displaying the results.
+func (link Link) PrettyPrintJSON() string {
```
jrperritt: What is the use-case for this? The `gophercloud.Link`s should be formatted if they are in the result and `gophercloud.Result.PrettyPrintJSON` is called. When would you need to format the links separately?

pratikmallya: @jrperritt that will format the entire result as JSON; but there are cases where we don't want all the result as JSON, only the links. e.g. https://github.com/rackspace/rack/pull/346/files#diff-89ed2b55dee1ba0cb9742aba85a8f6deR58

pratikmallya:
After discussing further, I agree we do not need this method

PR #499, opened 2015-10-03
Updating godocs descriptions for SecGroupRule

tahia-khan:
Documentation update

I noticed that some of the godoc method descriptions in the networking/v2/extensions/security/rules package (eg [Create]( https://godoc.org/github.com/rackspace/gophercloud/openstack/networking/v2/extensions/security/rules#Create)) were  referring to Security Groups instead of Security Group Rules. I'm just updating the function descriptions in those cases.


jrperritt:
Thank you :smile:  +2

PR #502, opened 2015-10-13
Adding 200 response code for POST request

goru97:
*What*:
Added 200 as ok response code for POST requests.

*Why*:
There is one Rackspace-metrics API (http://docs.rackspace.com/cmet/api/v1.0/cmet-devguide/content/POST_sendMetricsList_views_Query_Operations.html) that responds with 200 code against POST request.

*Is it Okay?*:
Yes, It's OK to send 200 against POST requests.
http://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html

### DIFF
--- a/provider_client.go
+++ b/provider_client.go
@@ -240,7 +240,7 @@ func defaultOkCodes(method string) []int {
 	case method == "GET":
 		return []int{200}
 	case method == "POST":
-		return []int{201, 202}
+		return []int{200, 201, 202}
 	case method == "PUT":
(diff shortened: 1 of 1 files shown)


jrperritt:
Instead, just override the expected response codes like here:
https://github.com/rackspace/gophercloud/blob/master/openstack/networking/v2/extensions/lbaas/pools/requests.go#L142

goru97:
@jrperritt That worked! Thanks very much :)

PR #503, opened 2015-10-13
[WIP] - Metrics Integration with GopherCloud.

goru97:
*What:*
Adding metrics support in gophercloud.

*Why:* 
On-demand from our customers.

*How:*
By writing code in GOLang.

*Additional Info:*
100% test coverage.


goru97:
@jrperritt Am I doing this the right way?

rgbkrk:
Really glad to see this coming along. The Travis test failure on go tip is unrelated (only on RackConnect).

jrperritt:
@goru97 Not quite, but you're definitely on the right track, and the unit test coverage looks fantastic. Have a look at the `List` function [here](https://github.com/rackspace/gophercloud/blob/master/openstack/networking/v2/networks/requests.go#L41) to see how we pass optional parameters to functions and then build the query string and HTTP headers. For example, the functions in `urls.go` should just return the base URL (sans query string).

goru97:
@jrperritt Thanks for your feedback. I'll work on it :)

goru97:
@jrperritt @rgbkrk Done some changes with the query builder; Also added the code for ingestion-api. Please take a look at your convenience.

rackspace/metrics/v2/ingestion/requests.go
```diff
@@ -0,0 +1,205 @@
...
+func SendMetrics(c *gophercloud.ServiceClient, metrics []MetricData) {
+	var res PostResult
+	reqBody := make([]map[string]interface{}, len(metrics))
+	for i := range metrics {
+		Metric := metrics[i]
+		reqBody[i] = map[string]interface{}{
+			"collectionTime": Metric.CollectionTime,
+			"ttlInSeconds":Metric.TtlInSeconds,
+			"metricValue": Metric.MetricValue,
+			"metricName": Metric.MetricName,
```
rgbkrk: A lot of these look like `struct`s you could define along with tags for what they serialize to in JSON. There's a lot of repetition here for fields.

jrperritt: Yeah, this is still not quite right. Have a look at [this method](https://github.com/rackspace/gophercloud/blob/master/openstack/networking/v2/networks/requests.go#L101) to see how we pass parameters to functions and then act on them. Notice how the `Create` function in the file above is very minimal.

goru97: @rgbkrk @jrperritt Sorry for replying late, got busy with some other stuff. Here's the gist of requests.go file that I wrote based on your feedback. See if I've done it correct this time. It's my first time with GO so I regret the inconvenience. https://gist.github.com/goru97/3814b41f9f4b1ecbd14e

rgbkrk: @goru97 Hey, it's ok. We're happy to help any way we can. Go can be tricky to get right at first (or second, or third... :wink:)

rgbkrk: I like the aggregate structs you've provided. Generally I'm happier when there's less `map[string]interface{}` and more declared and defined structs - it is a compiled language after all.

goru97: @rgbkrk Thanks! I'll go ahead and commit this new file. I am not sure about less map[string]interface{}. I am only using these to create request bodies out of defined structs, the way it is done in the Create function as mentioned by @jrperritt in his above comment. Are you saying there is a better way to do it using reflection and extracting tags from defined structs. I am confused now :stuck_out_tongue_winking_eye: I would really appreciate if you give me a gist after modifying a function a little bit so that I can get a better understanding.

sam-falvo: Whether or not map[string]interface{} is the best solution or not will depend on how that construct is actually used.  Gophercloud has changed significantly since I last maintained it, but since I was asked for help, I'd like to be brought up to speed on how you intend on using this map.  Why is it being produced?  What's consuming it?  Etc.

sam-falvo: One more question: concerning this code:
```go
reqBody[i] = map[string]interface{}{
	"collectionTime": Metric.CollectionTime,
	"ttlInSeconds":Metric.TtlInSeconds,
	"metricValue": Metric.MetricValue,
	"metricName": Metric.MetricName,
...etc...
```
Will this collection of JSON attributes change from within Gophercloud?  E.g., will there be protocol extensions which Gophercloud has to deal with that fundamentally alters which attributes appear inside this JSON object?

**If not**, then I'd agree that you should replace this with a `struct` with some JSON encoding attributes.  Look in the encoding/json package documentation for more details on how that's done.

**If so,** then I'm sorry @rgbkrk, but there's no hope.  It'll have to stay a map of interfaces because there's no way to statically predefine variant records in a JSON-encodable manner.  This is why I wanted to know more about the code.  Lots of variables and unknowns at this point.

goru97: Thanks @sam-falvo for helping out! I now understood how to directly decode/encode structure to map or vice-versa using "mapstructure". So for the fixed fields I'll be using that and will try to remove as many redundancies as possible. Thanks @rgbkrk for your patience :smiley:

rgbkrk: If you're directly setting attributes to send, you know the structure you're sending.

rgbkrk:
Yes, thank you so much for the structs. I'll leave @jrperritt to the rest.

goru97:
@rgbkrk Thank you for your patience and guidance :blush:

rackspace/client.go
```diff
@@ -212,3 +212,13 @@ func NewRackConnectV3(client *gophercloud.ProviderClient, eo gophercloud.Endpoin
 	}
 	return &gophercloud.ServiceClient{ProviderClient: client, Endpoint: url}, nil
 }
+
+// NewRackMetricsV2 creates a ServiceClient that may be used to access the v2 Rackspace metrics service.
+func NewRackMetricsV2(client *gophercloud.ProviderClient, eo gophercloud.EndpointOpts) (*gophercloud.ServiceClient, error) {
```
jrperritt: Since we're in the `rackspace` package, just `NewMetricsV2`.

sam-falvo:
One final, parting thought from me: make sure you run "go fmt" over your code.  I've noticed some odd code formatting which go fmt will automatically fix for you.

rackspace/client.go
goru97: Done!

goru97:
@jrperritt Do I need to write acceptance tests as well or is it considered done?

jrperritt:
@goru97 Yes, you'll need to write acceptance tests, but I'd recommend waiting until the API is set before doing that. I'll go through today and point out some more things in the PR that need changing.

rackspace/metrics/v2/doc.go
```diff
@@ -0,0 +1,9 @@
+//Rackspace Metrics is a software service that enables users to ingest,
+//condense, and query various kinds of metrics. Rackspace Metrics enables users to
+//quickly iterate through the data with different queries and set up dashboard metrics
+//for ongoing metrics monitoring. Rackspace Metrics is suitable for large-scale data collection
+//but does not sacrifice the query performance that is needed for interactive metrics exploration.
+
+// Package metrics allows Rackspace cloud accounts to leverage version 2 of
```
jrperritt: s/Package metrics/Package v2

goru97: Changed to package metrics; Similar to rackconnect.

rackspace/metrics/v2/query/requests.go
```diff
@@ -0,0 +1,140 @@
...
+	Until      int64    `q:"until"`
+	Points     int32    `q:"points"`
+	Resolution string   `q:"resolution"`
+	Select     []string `q:"select"`
+	Query      string   `q:"query"`
+	Tags       string   `q:"tags"`
+}
+
+// GetDataForListByPoints retrieve data against a list of metrics and number of points, for the specified tenant associated with RackspaceMetrics.
+func GetDataForListByPoints(c *gophercloud.ServiceClient, opts QueryParams, metrics ...string) (MetricListData, error) {
```
jrperritt: All functions should return some kind of `Result` and then extract the content via an `Extract` method in `results.go`.

rackspace/metrics/v2/query/results.go
```diff
@@ -0,0 +1,108 @@
...
+// MetricList is a part of MetricListData.
+type MetricList struct {
+	Unit   string  `mapstructure:"unit"`
+	Metric string  `mapstructure:"metric"`
+	Data   []Value `mapstructure:"data"`
+	Type   string  `mapstructure:"type"`
+}
+
+// []Metric is the result for metric search query.
+type Metric struct {
```
jrperritt: Where is this used?

rackspace/metrics/v2/query/results.go
```diff
@@ -0,0 +1,108 @@
...
+	gophercloud.Result
+}
+
+// MetricListData represents the result of query for multiple metrics.
+type MetricListData struct {
+	Metrics []MetricList `mapstructure:"metrics"`
+}
+
+// MetricList is a part of MetricListData.
+type MetricList struct {
```
jrperritt: Metric

rackspace/metrics/v2/query/results.go
```diff
@@ -0,0 +1,108 @@
...
+	Unit   string  `mapstructure:"unit"`
+	Metric string  `mapstructure:"metric"`
```
jrperritt: Name

rackspace/metrics/v2/query/results.go
```diff
@@ -0,0 +1,108 @@
...
+	Average   float64 `mapstructure:"average"`
+	Sum       float64 `mapstructure:"sum"`
+	Max       float64 `mapstructure:"max"`
+	Min       float64 `mapstructure:"min"`
+	Latest    float64 `mapstructure:"latest"`
+	Variance  float64 `mapstructure:"variance"`
+}
+
+//Metadata is the additional info about MetricData hence, it is a part of MetricData.
+type MetaData struct {
```
jrperritt: Metadata

rackspace/metrics/v2/query/results.go
```diff
@@ -0,0 +1,108 @@
...
+// []Metric is the result for metric search query.
+type Metric struct {
+	Metric string `mapstructure:"metric"`
+}
+
+// MetricListData represents the result of query for a metric.
+type MetricData struct {
+	Unit     string   `mapstructure:"unit"`
+	Values   []Value  `mapstructure:"values"`
+	MetaData MetaData `mapstructure:"metadata"`
```
jrperritt: Metadata

rackspace/metrics/v2/query/results.go
```diff
@@ -0,0 +1,108 @@
...
+//Rate is a part of Limits
+type Rate struct {
+	Limit []Limit `mapstructure:"limit"`
+	Regex string  `mapstructure:"regex"`
+	URI   string  `mapstructure:"uri"`
+}
+
+//Limit is a part of Rate
+type Limit struct {
+	Next_Available string `mapstructure:"next-available"`
```
jrperritt: No underscores

rackspace/metrics/v2/query/results.go
```diff
@@ -0,0 +1,108 @@
...
+	Max       float64 `mapstructure:"max"`
+	Min       float64 `mapstructure:"min"`
+	Latest    float64 `mapstructure:"latest"`
+	Variance  float64 `mapstructure:"variance"`
+}
+
+//Metadata is the additional info about MetricData hence, it is a part of MetricData.
+type MetaData struct {
+	Limit     string `mapstructure:"limit"`
+	Next_Href string `mapstructure:"next_href"`
```
jrperritt: No underscores

rackspace/metrics/v2/query/results.go
```diff
@@ -0,0 +1,108 @@
...
+}
+
+func (gr GetResult) GetValues() ([]Value, error) {
+	var res []Value
+	b := gr.Body.(map[string]interface{})
+	err := mapstructure.Decode(b["values"], &res)
+	return res, err
+}
+
+func (gr GetResult) GetMetricsData() (MetricData, error) {
```
jrperritt: `Extract`

rackspace/metrics/v2/query/results.go
```diff
@@ -0,0 +1,108 @@
...
+// ExtractMetadata is a function that takes a GetResult (of type *http.Response)
+// and returns the custom metatdata associated with the account.
+func (gr GetResult) ExtractMetadata() (MetaData, error) {
+	var res MetaData
+	b := gr.Body.(map[string]interface{})
+	err := mapstructure.Decode(b["metadata"], &res)
+	return res, err
+}
+
+func (gr GetResult) GetValues() ([]Value, error) {
```
jrperritt: `ExtractValues`

jrperritt:
@goru97 I mentioned a few basic changes to make. Again, I'd encourage you go through the other Gophercloud packages and modify this PR to follow the style/naming/logic conventions shown there. Unrelated to that, I'm not crazy about the package layout here. That's no fault of yours; you've organized it as it's shown in the docs. But it doesn't read particularly well when being coded, and therefore doesn't conform to the Gophercloud style. Off the top of my head, instead of `query` and `ingestion`, I'm leaning toward 4 subpackages: `metrics`, `counters`, `gauges`, `timers` with the appropriate functions/objects in each. With that layout, we'd have `metrics.List`, `metrics.ListAvailable`, `counters.Get`, `metrics.Limits`, `timers.Send` etc. Maybe `SendAggregated`/`GetAggregated`, I'm unsure if there's a difference.

goru97:
@jrperritt I understand your concern about the package structure but changing it to 4 sub packages will make it worse because there are two different endpoints for metrics (one for ingest and other for query) and that's why for query functions, one has to pass ServiceClient for query endpoint and similarly for ingestion. But if we change it to metrics, counters, gauges... one has to pass ServiceClient for query endpoint while doing e.g. counters.send() and ServiceClient for ingest while doing counters.get(), which will only increase the complexity and also the confusion.

rackspace/metrics/v2/query/results.go
goru97: Here: https://github.com/goru97/gophercloud/blob/1b6829d25cfde569c881e3313813eb7c9238769a/rackspace/metrics/v2/query/requests.go#L101

jrperritt:
@goru97 You have a single `NewMetricsv2` function for getting a Cloud Metrics client. What would a user provide differently to that function to get the different endpoints? Do you have an example of what the Cloud Metrics entry looks like in the catalog?

goru97:
@jrperritt Here's the gist for making service clients. Notice the service name passed in the endpointOpts.
https://gist.github.com/goru97/77da071cf930230d4cd1

jrperritt:
If there are 2 different endpoints, there should be 2 different clients and 2 different packages beneath the `rackspace` package.

goru97:
@jrperritt Did you mean one  metricsQuery and other metricsIngest?!!

jrperritt:
Yes, but all lowercase.

jrperritt:
Closing due to lack of activity

Issue #498, opened 2015-10-01
Infinite recursion with AllowReauth


marcin-ptaszynski:
Hi guys,

It's possible to cause client to run infinite recursion (or until it hits file descriptor limit) during [Reauth Attempt](https://github.com/rackspace/gophercloud/blob/d86f9582b08d56dc9422e3a5d9f120b117487427/provider_client.go#L187-L195), it receives 401 response. Below code uses v2 identity API, but the same happens when we use `github.com/rackspace/gophercloud/openstack/identity/v3/tokens` `Get()`

```
package main

import (
	"bufio"
	"encoding/json"
	"fmt"
	"os"

	"github.com/rackspace/gophercloud"
	"github.com/rackspace/gophercloud/openstack"
)

func main() {
	opts := gophercloud.AuthOptions{
		IdentityEndpoint: "http://keystone-host:5000/v2.0",
		Username:         "user",
		Password:         "pass",
		TenantName:       "admin",
		AllowReauth:      true,
	}

	providerClient, err := openstack.AuthenticatedClient(opts)
	if err != nil {
		panic(err)
	}

	serviceClient := openstack.NewIdentityV2(providerClient)

	stdin := bufio.NewReader(os.Stdin)

	for {
		fmt.Println("<<<<")

		byteToken, _, err := stdin.ReadLine()
		if err != nil {
			panic(err)
		}
		token := string(byteToken)
		fmt.Printf(".... \"%s\"\n", token)

		var result interface{}
		_, err = serviceClient.Get(serviceClient.ServiceURL("tokens", token), &result, &gophercloud.RequestOpts{
			OkCodes: []int{200, 203},
		})
		if err != nil {
			fmt.Println("error: ", err)
		}

		json, _ := json.MarshalIndent(result, "", "  ")
		fmt.Println(">>>>", string(json))
	}
}
```

In order to reproduce
* change user password in the middle of execution
* write to stdin token from different domain (v3), etc

In each case client takes up all fd and doesn't close them even after error
`(Error trying to re-authenticate: * a lot)Error trying to re-authenticate: Error trying to re-authenticate: Error trying to re-authenticate: Error trying to re-authenticate: Error trying to re-authenticate: Error trying to re-authenticate: Error trying to re-authenticate: Error trying to re-authenticate: Error trying to re-authenticate: Error trying to re-authenticate: Post http://10.33.5.214:5000/v2.0/tokens: dial tcp 10.33.5.214:5000: too many open files`
 is returned.

[...]

PR #508, opened 2015-10-27
add validate user's token method for v2 and bug fix for reauth

chaolou:
1. Now keystone v2's identity has only a `Create()` method, but we need to use v2's api for dynamic auth k8s, so I add a `Get()` method.
2. The `reauth` in client.go always fail because it forget to clear the expire X-Auth-Token and should reset after reauth successful,  also the recursive `ReauthFunc` will never end if the admin reauth fail, so it need to clear the `ReauthFunc` pointer before executing `ReauthFunc` and restore after that.

### DIFF
--- a/openstack/identity/v2/tokens/requests.go
+++ b/openstack/identity/v2/tokens/requests.go
@@ -91,12 +91,12 @@ func Create(client *gophercloud.ServiceClient, auth AuthOptionsBuilder) CreateRe
 
 // Validates and retrieves information for user's token.
 func Get(client *gophercloud.ServiceClient, token string) GetResult {
-    var result GetResult
-    _, result.Err = client.Get(CreateGetURL(client, token), &result.Body, &gophercloud.RequestOpts{
-        OkCodes: []int{200, 203},
-    })
-    if result.Err != nil {
-        return result
-    }
-    return result
+	var result GetResult
+	_, result.Err = client.Get(GetURL(client, token), &result.Body, &gophercloud.RequestOpts{
+		OkCodes: []int{200, 203},
+	})
+	if result.Err != nil {
+		return result
+	}
+	return result
 }
--- a/openstack/identity/v2/tokens/requests.go
+++ b/openstack/identity/v2/tokens/requests.go
@@ -88,3 +88,15 @@ func Create(client *gophercloud.ServiceClient, auth AuthOptionsBuilder) CreateRe
 	})
 	return result
 }
+
+// Validates and retrieves information for user's token.
+func Get(client *gophercloud.ServiceClient, token string) GetResult {
+    var result GetResult
+    _, result.Err = client.Get(CreateGetURL(client, token), &result.Body, &gophercloud.RequestOpts{
+        OkCodes: []int{200, 203},
+    })
+    if result.Err != nil {
+        return result
+    }
+    return result
+}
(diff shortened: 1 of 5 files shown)


openstack/identity/v2/tokens/urls.go
```diff
@@ -6,3 +6,8 @@ import "github.com/rackspace/gophercloud"
 func CreateURL(client *gophercloud.ServiceClient) string {
 	return client.ServiceURL("tokens")
 }
+
+// CreateGetURL generates the URL used to Validate Tokens.
+func CreateGetURL(client *gophercloud.ServiceClient, token string) string {
```
jrperritt: We're naming these functions based on their corresponding requests, so just `GetURL`.

openstack/identity/v2/tokens/results.go
```diff
@@ -74,6 +74,11 @@ type CreateResult struct {
 	gophercloud.Result
 }
 
+// GetResult is the deferred response from a Get call.
+type GetResult struct {
```
jrperritt: Is there a corresponding `Extract` method for this?

jrperritt:
@zyren88 The reauth stuff (like limiting the number of retries) should be handled in your app code by passing in a custom client: https://github.com/rackspace/gophercloud/blob/master/provider_client.go#L62

The change you made for resetting the token in the reauth function we can keep though.

zyren88:
@jrperritt  the existing reauth logic could trap into an infinite recursive loop in some cases, such as user has change their password of keystone. It's a bug which must be fixed from my perspective, instead of an optional retry limits. 

Anyway, I will rework my PR according to your suggestion. However, I highly recommend to just reauth one time, since there is no need to retry reauth several times when get a `401 ` http code when call `ReauthFunc`(not a network problem, but a `Unauthorized` error!).

jrperritt:
I completely understand what you're saying, and it's an edge case that some users (maybe you) will have to deal with. It is easily handled by passing in a custom `http.RoundTripper, and you can see an example of how to do it by following the link in my comment here: https://github.com/rackspace/gophercloud/issues/498#issuecomment-144839363. More than one reauth attempt is sometimes necessary; I've encountered it several times.

openstack/identity/v2/tokens/requests.go
```diff
@@ -88,3 +88,15 @@ func Create(client *gophercloud.ServiceClient, auth AuthOptionsBuilder) CreateRe
...
 	return result
 }
+
+// Validates and retrieves information for user's token.
+func Get(client *gophercloud.ServiceClient, token string) GetResult {
+	var result GetResult
+	_, result.Err = client.Get(GetURL(client, token), &result.Body, &gophercloud.RequestOpts{
+		OkCodes: []int{200, 203},
+	})
+	if result.Err != nil {
```
jrperritt: We don't need to check the error here; it'll get done in `Extract`.

openstack/identity/v2/tokens/results.go
```diff
@@ -23,6 +23,10 @@ type Token struct {
 
 	// Tenant provides information about the tenant to which this token grants access.
 	Tenant tenants.Tenant
+
+	// the owner user of token
+	UserName string
+	UserID   string
```
jrperritt: I'm not sure we want these as part of the `Token`. Maybe instead we have an `ExtractUser` method?

jrperritt:
In addition to the above considerations, we should probably include unit and acceptance tests.

chaolou:
The new modification move to https://github.com/rackspace/gophercloud/pull/514, so I close this

PR #509, opened 2015-10-31
[rfr] Remove imageName and flavorName from Boot From Volume

jtopjian:
As of Liberty, passing imageName and flavorName causes Nova to return
a 400 error reporting that those fields are not allowed to be sent.


jtopjian:
@jrperritt back with another BFV issue :smile: 

It's pretty easy to reproduce since the acceptance test fails on Liberty without this patch.

Can you do a quick review and let me know if there's a better way to do this?

jrperritt:
Ah, yep. Hmm, off the top of my head, I'm not sure why I put that logic in `Create` and not `ToServerCreateMap`.

+2

jtopjian:
Thanks you, sir!

PR #514, opened 2015-11-30
Bug fix reauth and add extract user from token

chaolou:
Modifed according to @jrperritt suggestions. the old pull request see https://github.com/rackspace/gophercloud/pull/508 which has beed

1. Clear the auth-token in the reauth function.

2. Add extract user from validate token's repsonse including unit and acceptance tests.


jrperritt:
Thank you for the changes. This looks good. +2

PR #515, opened 2015-12-14
[rfr] multi ephemeral support

jtopjian:
This commit adds the ability to specify multiple ephemeral disks through the
bootfromvolume extension.


jtopjian:
FYI: The acceptance test will fail if the flavor does not have ephemeral space configured. 

More details about this feature: https://github.com/hashicorp/terraform/issues/4042

jrperritt:
Looks good. +2

jtopjian:
Thank you!

PR #518, opened 2016-01-07
Fixed problem re-authenticating during request

clintonskitson:
This commit fixes the nested re-auth logic where it was not
properly returning the response and causing a nil body to
be attempted to be read.

The error being returned was "http invalid read on closed 
request body".


jrperritt:
This seems reasonable to me. +2

PR #519, opened 2016-01-08
[rfr] Allow all routes to be removed from a Neutron router

timbyr:
Remove omitempty tag so an empty Routes array will be passed in JSON
request and all routes can be removed.


jrperritt:
Sounds fine to me. +2

PR #520, opened 2016-01-14
Use better type assertions for Trove/CDB results

jamiehannaford:
Inline with https://github.com/rackspace/gophercloud/blob/master/openstack/orchestration/v1/stackevents/results.go#L111


jrperritt:
+2

Issue #521, opened 2016-01-18
ICMP Type / Port Can't be Zero


jtopjian:
Per https://github.com/hashicorp/terraform/issues/4716, the From Port and To Port should accept values of 0 when the IP Protocol is "ICMP".

jtopjian:
Fixed in #522

PR #522, opened 2016-01-19
[rfr] From Port and To Port should accept values of 0 

deniszh:
…when the IP Protocol is "ICMP".
Fixing https://github.com/rackspace/gophercloud/issues/521

### DIFF
--- a/openstack/compute/v2/extensions/secgroups/requests.go
+++ b/openstack/compute/v2/extensions/secgroups/requests.go
@@ -2,7 +2,7 @@ package secgroups
 
 import (
 	"errors"
-    "strings"
+	"strings"
 
 	"github.com/rackspace/gophercloud"
 	"github.com/rackspace/gophercloud/pagination"
--- a/openstack/compute/v2/extensions/secgroups/requests.go
+++ b/openstack/compute/v2/extensions/secgroups/requests.go
@@ -2,6 +2,7 @@ package secgroups
 
 import (
 	"errors"
+    "strings"
 
 	"github.com/rackspace/gophercloud"
 	"github.com/rackspace/gophercloud/pagination"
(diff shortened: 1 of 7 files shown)


deniszh:
Not sure do I need to test that. Tried to hack existing tests but failed.

openstack/compute/v2/extensions/secgroups/requests.go
```diff
@@ -2,6 +2,7 @@ package secgroups
 
 import (
 	"errors"
+    "strings"
```
jtopjian: spacing issue.

deniszh:
fixed, thanks!

deniszh:
Yay, we have a test now! Thanks, @jtopjian!

jtopjian:
np :smile: 

Just a heads up to the Gophercloud team: there's something preventing security group tests from succeeding. The error seen (without this patch) is:

```
convenience.go:34: Failure in secgroup_test.go, line 160: unexpected error "EOF"
```

Discarding the error check makes everything work.

Also, this patch is required because it looks as though the Contrail Neutron plugin is mangling ICMP rules in OpenStack. Even though `-1/-1` is used as input, `0/65535` is returned.

jtopjian:
@deniszh Can you append an [rfr] tag to the beginning of this PR? That notifies the gophercloud team to review the PR.

@jrperritt if you have time, can you give this a quick review and let us know if it's something that can be merged quickly or needs more thought? All Contrail-based Neutron environments are affected by this.

deniszh:
@jtopjian - like that? I can't put any label to it

jtopjian:
Yup, just like that. :)

jrperritt:
This looks good but needs a unit test.

deniszh:
@jrperritt: it already has test from @jtopjian - but if that is not enough - what test is needed then?

jrperritt:
It has an acceptance test; it needs a unit test. See here:
https://github.com/deniszh/gophercloud/blob/master/openstack/compute/v2/extensions/defsecrules/requests_test.go
and
https://github.com/deniszh/gophercloud/blob/master/openstack/compute/v2/extensions/secgroups/requests_test.go

jtopjian:
@deniszh I've just opened a PR to your local repo with the applicable unit tests.

deniszh:
Wow, @jtopjian, thanks a lot, again!
Was completely busy with other stuff...

Build status is green now.
@jrperritt - does this looks good?

jrperritt:
Thank you @jtopjian and @deniszh . +2

PR #525, opened 2016-01-28
Add support for distributed routers

Fodoj:
Even though it is not referenced in API ref, it is official feature, available in neutron cli, for example: http://docs.openstack.org/cli-reference/neutron.html


jtopjian:
If you run `neutron --debug router-create` with the `--distributed` flag, are you able to see the "distributed" info in the request and response? If so, I think it would be helpful to update the JSON fixtures in the `requests_tests.go` file.

Fodoj:
@jtopjian added, thanks

jrperritt:
+2

PR #528, opened 2016-02-02
Fix keystone v3 token auth

orivej:
Currently keystone v3 client ignores TokenID in AuthOptions and the code like this
```go
provider, err := openstack.AuthenticatedClient(gophercloud.AuthOptions{
	IdentityEndpoint: "https://.../v3/",
	TokenID:          "abcd...1234",
})
```
fails with an error `You must provide a password to authenticate` because `options.TokenID` is never checked. (keystone v2 client checks it in `ToTokenCreateMap`.)  This patch lets authentication succeed.

I am new to the codebase and do not understand if this issue should be fixed like this, and I would also appreciate help in writing a test.

### DIFF
--- a/openstack/auth_env.go
+++ b/openstack/auth_env.go
@@ -13,14 +13,15 @@ var nilOptions = gophercloud.AuthOptions{}
 // environment variables, respectively, remain undefined.  See the AuthOptions() function for more details.
 var (
 	ErrNoAuthURL  = fmt.Errorf("Environment variable OS_AUTH_URL needs to be set.")
-	ErrNoUsername = fmt.Errorf("Environment variable OS_USERNAME needs to be set.")
-	ErrNoPassword = fmt.Errorf("Environment variable OS_PASSWORD needs to be set.")
+	ErrNoUsername = fmt.Errorf("Environment variable OS_USERNAME, OS_USERID, or OS_TOKEN needs to be set.")
+	ErrNoPassword = fmt.Errorf("Environment variable OS_PASSWORD or OS_TOKEN needs to be set.")
 )
 
-// AuthOptions fills out an identity.AuthOptions structure with the settings found on the various OpenStack
-// OS_* environment variables.  The following variables provide sources of truth: OS_AUTH_URL, OS_USERNAME,
-// OS_PASSWORD, OS_TENANT_ID, and OS_TENANT_NAME.  Of these, OS_USERNAME, OS_PASSWORD, and OS_AUTH_URL must
-// have settings, or an error will result.  OS_TENANT_ID and OS_TENANT_NAME are optional.
+// AuthOptionsFromEnv fills out an AuthOptions structure from the environment
+// variables: OS_AUTH_URL, OS_USERNAME, OS_USERID, OS_PASSWORD, OS_TENANT_ID,
+// OS_TENANT_NAME, OS_DOMAIN_ID, OS_DOMAIN_NAME, OS_TOKEN.  It checks that
+// (1) OS_AUTH_URL is set, (2) OS_USERNAME, OS_USERID, or OS_TOKEN is set,
+// (3) OS_PASSWORD or OS_TOKEN is set.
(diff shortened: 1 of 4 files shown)


orivej:
In the second patch I enabled scoped token auth as in:
```go
provider, err := openstack.AuthenticatedClient(gophercloud.AuthOptions{
	IdentityEndpoint: "https://.../v3/",
	TokenID:          "abcd...1234",
	DomainID:         "default",
	TenantName:       tenant,
})
```
which is documented at http://developer.openstack.org/api-ref-identity-v3.html#authenticateTokenScoped

jrperritt:
The second patch looks good.

For the first, v3 auth sets the Token [here](https://github.com/rackspace/gophercloud/blob/master/openstack/identity/v3/tokens/requests.go#L106). Instead of `c.TokenID` I think changing it to `opts.TokenID` would do it, no? Also, line 87 would need the same change.

orivej:
> Instead of c.TokenID I think changing it to opts.TokenID would do it, no?

[Create](https://github.com/rackspace/gophercloud/blob/231898eeeb3f1efabaf953757babcdb84298fbe8/openstack/identity/v3/tokens/requests.go#L23) seems to be supposed to be able to reauthenticate existing clients, using token cached in `c.TokenID`. This is tested by [authTokenPost](https://github.com/rackspace/gophercloud/blob/231898eeeb3f1efabaf953757babcdb84298fbe8/openstack/identity/v3/tokens/requests_test.go#L20) and other tests.

You can think of `client.TokenID = options.TokenID` as prepopulating token that would be returned to the client if it authenticated with login and password.

Although with my patches all tests pass, I think we should replace proposed
```
client.TokenID = options.TokenID
```
with
```
if len(options.TokenID) != 0 {
	client.TokenID = options.TokenID
}
```
and move it out of the switch.

orivej:
Disregard the suggestion of moving it out of the switch: v2 client already processes `options.TokenID` and seems to ignore `client.TokenID`.

orivej:
I moved the original proposal to https://github.com/orivej/gophercloud/tree/openstack-identity-v3-pass-tokenid%401 and improved the first patch by relocating token assignment into `v3/tokens/request.go`.

orivej:
Rebased onto master, combined into one commit, and added a test that covers the addition to `requests.go`. I do not know why coveralls reports that it is not covered.

orivej:
Updated ErrMissingPassword message.

orivej:
The coverage decreases because `github.com/rackspace/gophercloud/openstack/identity/v3/tokens` tests do not cover the addition to `requests.go`, but `github.com/rackspace/gophercloud/openstack` do cover it.

orivej:
I have moved the new test into `request_test.go`, and split the patch back in two.

openstack/auth_env.go
```diff
@@ -30,16 +31,17 @@ func AuthOptionsFromEnv() (gophercloud.AuthOptions, error) {
 	tenantName := os.Getenv("OS_TENANT_NAME")
 	domainID := os.Getenv("OS_DOMAIN_ID")
 	domainName := os.Getenv("OS_DOMAIN_NAME")
+	tokenID := os.Getenv("OS_TOKEN")
```
jrperritt: Can you help me understand the use-case for this? Since tokens are presumably ephemeral, why would we want to load them from an environment variable? And if a user has set `OS_TOKEN`, what then when the token inevitably expires?

orivej: https://specs.openstack.org/openstack/keystone-specs/api/v3/identity-api-v3-os-oauth1-ext.html says that OAuth tokens 

> Provide the ability for identity users to delegate roles to third party consumers via the OAuth 1.0a specification. Requires v3.0+ of the Identity API. An OAuth-derived token will provide a means of acting on behalf of the authorizing user.

They can be revoked by the user or expire (which is optional), after which they are meant to no longer provide authorization to perform on behalf of the user.

jrperritt:
LGTM. +2

PR #529, opened 2016-02-03
[rfr] Fix reauth in v3 tokens

### LINKED ISSUE #523: Cannot reauth in identity/v3/tokens package
# Summary
Gophercloud cannot reauth properly when performing any request in identity/v3/tokens package.

# Steps to reproduce
1. Set keystone token expiration time to 10 seconds (not necessary, but helpful).
2. Create ServiceClient with RoundTripper which limits retry times to 3. Without mentioned RoundTripper, gophercloud will enter endless recursion.
3. Wait till token expires (10 second in my case).
4. Call one of tokens.{Create,Get,Revoke,Validate}.
5. Receive 401 instead of expected response.

## Code used to reproduce
```go
package main

import (
	"fmt"
	"net/http"
	"time"

	"github.com/rackspace/gophercloud/openstack"
	"github.com/rackspace/gophercloud/openstack/identity/v3/tokens"
)

//RoundTripper limits number of Reauth attempts
type RoundTripper struct {
	rt                http.RoundTripper
	numReauthAttempts int
	maxReauthAttempts int
}

func newHTTPClient() http.Client {
	return http.Client{
		Transport: &RoundTripper{
			rt:                http.DefaultTransport,
			maxReauthAttempts: 3,
		},
	}
}

//RoundTrip limits number of Reauth attempts
func (lrt *RoundTripper) RoundTrip(request *http.Request) (*http.Response, error) {
	var err error

	response, err := lrt.rt.RoundTrip(request)
	if response == nil {
		return nil, err
	}

	if response.StatusCode == http.StatusUnauthorized {
		if lrt.numReauthAttempts == lrt.maxReauthAttempts {
			return response, fmt.Errorf("Failed to reauthenticate"+
				"to keystone with %d attempts", lrt.maxReauthAttempts)
		}
		lrt.numReauthAttempts++
	}
	return response, err
}

func main() {
	opts, err := openstack.AuthOptionsFromEnv()
	if err != nil {
		fmt.Printf("%s\n", err.Error())
	}
	opts.AllowReauth = true
	provider, err := openstack.AuthenticatedClient(opts)
	if err != nil {
		fmt.Printf("%s\n", err.Error())
	}
	client := openstack.NewIdentityV3(provider)
	client.HTTPClient = newHTTPClient()

	// Wait till token will expire
	time.Sleep(10 * time.Second)

	// This should return 404
	tokens.Get(client, client.TokenID).Extract()
	if err != nil {
		fmt.Printf("%s\n", err.Error())
	}
}
```

# Cause
Each ```RequestOpts.MoreHeaders``` is set to ```subjectTokenHeaders(...)```, which returns following map:
```go
map[string]string{
	"X-Auth-Token": "token-before-reauth",
	"X-Subject-Token": "some-token",
}
```

When [```Request```](https://github.com/rackspace/gophercloud/blob/master/provider_client.go#L126) function is called, "X-Auth-Token" header is set in [line 163](https://github.com/rackspace/gophercloud/blob/master/provider_client.go#L163), after that in [line 170](https://github.com/rackspace/gophercloud/blob/master/provider_client.go#L170) `options.MoreHeaders` are added to request, **which causes "X-Auth-Token" override** with incorrect one provided in `options.MoreHeaders`.

# Solutions
## Lazy one
Switch order of headers adding in [```Request```](https://github.com/rackspace/gophercloud/blob/master/provider_client.go#L126).

## Right(?) one
Change [`subjectTokenHeaders`](https://github.com/rackspace/gophercloud/blob/master/openstack/identity/v3/tokens/requests.go#L17) not to return "X-Auth-Token".

And here come my questions:
1. @smashwilson, Is there any reason why we should keep "X-Auth-Token" in map returned by `subjectTokenHeaders`? If no I will implement second solution.
2. @jrperritt, Is "X-Auth-Token" override possibility a desired functionality?

kwapik:
This fixes #523.

### DIFF
--- a/openstack/identity/v3/tokens/requests.go
+++ b/openstack/identity/v3/tokens/requests.go
@@ -15,9 +15,9 @@ type Scope struct {
 }
 
 func subjectTokenHeaders(c *gophercloud.ServiceClient, subjectToken string) map[string]string {
-	h := c.AuthenticatedHeaders()
-	h["X-Subject-Token"] = subjectToken
-	return h
+	return map[string]string{
+		"X-Subject-Token": subjectToken,
+	}
 }
 
 // Create authenticates and either generates a new token, or changes the Scope of an existing token.


jrperritt:
+2

PR #531, opened 2016-02-10
Support reauth on identity endpoints with a base path

alexcern:
Fix re-authentication to use the correct endpoint when the identity endpoint has a base path (e.g. https://keystone.example.com/somepath/v2.0)

So far the reauth function ignores the endpoint path and ends up using the wrong endpoint (in the example above, it will try to reauth against https://keystone.example.com/v2.0)


jrperritt:
Ah, yep. Thank you for filing this. +2

PR #532, opened 2016-02-13
Updated docs to provide information about tags needed to run tests.

rickard-von-essen:
Just a small improvement that would have saved me some confusion.


jrperritt:
Ah, yes, thank you :) +2

PR #535, opened 2016-02-16
Fix for issue#533

marcin-krolik:
Fixed call for list of available API versions for Cinder.

### DIFF
--- a/openstack/blockstorage/v1/apiversions/urls.go
+++ b/openstack/blockstorage/v1/apiversions/urls.go
@@ -2,6 +2,7 @@ package apiversions
 
 import (
 	"strings"
+	"net/url"
 
 	"github.com/rackspace/gophercloud"
 )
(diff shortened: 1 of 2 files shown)


openstack/blockstorage/v1/apiversions/urls.go
```diff
@@ -11,5 +12,7 @@ func getURL(c *gophercloud.ServiceClient, version string) string {
 }
 
 func listURL(c *gophercloud.ServiceClient) string {
-	return c.ServiceURL("")
+	u, _ := url.Parse(c.ServiceURL(""))
+	u.Path = "/"
+	return u.String()
```
jrperritt: Try just returning `c.Endpoint` such as [here](https://github.com/rackspace/gophercloud/blob/master/openstack/orchestration/v1/apiversions/urls.go#L6) and [here](https://github.com/rackspace/gophercloud/blob/master/openstack/networking/v2/apiversions/urls.go#L10) and see if that works.

marcin-krolik: Problem is ```c.Endpoint``` contains full service URL with path included
like ```http://public.fuel.local:8776/v2/d98e06adf5db49ad9f372625cad7840b/``` 
while for list of avaiable API version one needs to call root path 
like ```http://public.fuel.local:8776/```

jrperritt: Ah, I see now in the service catalog.

jrperritt:
LGTM. +2

PR #536, opened 2016-02-18
[WIP] Add missing attributes to GatewayInfo

rfc1459:
The `GatewayInfo` struct for a Neutron router is missing a few attributes returned by the Neutron routers API:
- `enable_snat` - which controls whether a given router performs SNAT for instances not bound to a floating IP
- `external_fixed_ips` - a list of subnet id/IP address pairs for each external address of this network

Given that `GatewayInfo` is used as part of the public API of gophercloud for Neutron routers, I slightly altered the Create and Update functions to discard the new attributes so that there would be no API breakage. I'm not really proud of the way it's done, though, but I didn't really want to introduce a `GatewayInfoOpts` struct for Create and Update at this point since that would break the public API.

TL;DR: feedback welcome :-)


rfc1459:
No longer relevant for me, closing

PR #538, opened 2016-02-23
Extracting User and Tenant details during get token details api call

uruddarraju:


jrperritt:
Can we instead do something like this from `identity v2`: https://github.com/uruddarraju/gophercloud/blob/master/openstack/identity/v2/tokens/results.go#L153 ?

PR #542, opened 2016-03-04
[rfr] os-quota-set GET api

dangogh:
still quite raw,  but feedback is most welcome.


openstack/compute/v2/extensions/quotasets/fixtures.go
```diff
@@ -0,0 +1,59 @@
...
+      "keypairs" : 10,
+      "injected_file_path_bytes" : 255
+   }
+}
+`
+
+const FirstTenantID = "555544443333222211110000ffffeeee"
+
+// FirstQuotaset is the first result in ListOutput.
+var FirstQuota = Quotaset{
```
jrperritt: `FirstQuotaSet`

openstack/compute/v2/extensions/quotasets/results.go
```diff
@@ -0,0 +1,105 @@
+package quotasets
+
+import (
+	"github.com/mitchellh/mapstructure"
+	"github.com/rackspace/gophercloud"
+	"github.com/rackspace/gophercloud/pagination"
+)
+
+// Quotaset is a set of operational limits that allow for control of compute usage.
+const sample = `
```
jrperritt: Maybe we could refactor this as the response for a unit test? At the least, let's move it to `fixtures.go`.

jrperritt:
This looks really good. Just 2 minor changes at first glance.

openstack/compute/v2/extensions/quotasets/fixtures.go
dangogh: so should it be QuotaSet throughout?

jrperritt: The type being called `QuotaSet` is fine. I was merely noting that in `requests_test.go`, you're referencing a variable named `FirstQuotaSet`, which doesn't exist and is causing the unit test to fail. Renaming this variable to `FirstQuotaSet` would likely fix that.

dangogh:
I think this is complete..   Please let me know if anything's missing.  I'll work on PUT/POST, etc as I can..

acceptance/openstack/compute/v2/quotaset_test.go
```diff
@@ -0,0 +1,60 @@
...
+	t.Logf("                   key_pairs=[%d]\n", quotaset.KeyPairs)
+	t.Logf("              metadata_items=[%d]\n", quotaset.MetadataItems)
+	t.Logf("             security_groups=[%d]\n", quotaset.SecurityGroups)
+	t.Logf("        security_group_rules=[%d]\n", quotaset.SecurityGroupRules)
+	t.Logf("                   fixed_ips=[%d]\n", quotaset.FixedIps)
+	t.Logf("                floating_ips=[%d]\n", quotaset.FloatingIps)
+	t.Logf(" injected_file_content_bytes=[%d]\n", quotaset.InjectedFileContentBytes)
+	t.Logf("    injected_file_path_bytes=[%d]\n", quotaset.InjectedFilePathBytes)
+	t.Logf("              injected_files=[%d]\n", quotaset.InjectedFiles)
+
```
jrperritt: For reference, you can format structs and maps with 
```go
t.Logf("quotaset: %+v", quotaset)
```
to print the object with its fields, but what you have now looks a lot nicer than that would.

jrperritt:
This looks great. +2

PR #545, opened 2016-03-23
[rfr] Add fixtures build tag to all fixtures.go files

timbyr:
Ensure testing dependencies from fixtures don't leak into builds.


jrperritt:
Ah, thank you :) +2

PR #546, opened 2016-03-24
[WIP] Add Rackspace Auto Scale support

### LINKED ISSUE #418: Rackspace Autoscale
Reference: http://docs.rackspace.com/cas/api/v1.0/autoscale-devguide/content/Overview.html

bison:
Adds support for Rackspace Auto Scale.

Docs: https://developer.rackspace.com/docs/autoscale/v1/developer-guide/

Fixes #418

### DIFF
--- a/rackspace/autoscale/v1/launch/fixtures.go
+++ b/rackspace/autoscale/v1/launch/fixtures.go
@@ -1,0 +1,89 @@
+// +build fixtures
+
+package launch
+
+import (
+	"fmt"
+	"net/http"
+	"testing"
+
+	th "github.com/rackspace/gophercloud/testhelper"
+	"github.com/rackspace/gophercloud/testhelper/client"
+)
+
+// LaunchGetBody contains the canned body of a launch.Get response.
+const LaunchGetBody = `
+{
+  "launchConfiguration": {
+    "args": {
+      "loadBalancers": [
+        {
+          "port": 443,
+          "loadBalancerId": 123456
+        }
+      ],
+      "server": {
+        "name": "server",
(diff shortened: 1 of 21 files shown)


rackspace/autoscale/v1/launch/fixtures.go
```diff
@@ -0,0 +1,89 @@
...
+				"foo": "bar",
+			},
+		},
+		LoadBalancers: []map[string]interface{}{
+			{
+				"port":           443,
+				"loadBalancerId": 123456,
+			},
+		},
+		DrainingTimeout: 0, // TODO: This is optional, is this right?
```
bison: This is optional, and may not be in the response when requesting a launch configuration -- zero is actually not a valid value. Should I ditch the `Args` type and have that just be a `map[string]interface{}`?

jrperritt: I think you've got the right idea here. We want the fields to be as specific as possible, so for some things we have no choice but `map[string]interface{}` but we want that to be  a last resort. If you don't want the `DrainingTimeout` field to show up if it's empty, you can use the [`encoding/json`](https://golang.org/pkg/encoding/json/#Unmarshal) package (instead of the `mapstructure` package) to unmarshal the JSON into the `Configuration` and use `json:"draining_timeout,omitempty"` for the field's struct tag.

bison:
So, all the bits for webhooks are done, and the diff is already pretty big. Would it be easier if I close this and open a separate PR for each package (webhooks, policies, launch, and groups) as I get to those, or do you prefer one giant PR?

jrperritt:
Smaller PRs would be appreciated.

bison:
Cool. Closing this.

PR #548, opened 2016-03-29
[RFR] Rackspace Auto Scale: webhooks

bison:
First steps toward supporting Rackspace Auto Scale. Adds a package for [webhook](https://developer.rackspace.com/docs/autoscale/v1/developer-guide/#document-api-operations/autoscale-webhooks) resources.

### DIFF
--- a/rackspace/autoscale/v1/webhooks/requests.go
+++ b/rackspace/autoscale/v1/webhooks/requests.go
@@ -1,0 +1,17 @@
+package webhooks
+
+import (
+	"github.com/rackspace/gophercloud"
+	"github.com/rackspace/gophercloud/pagination"
+)
+
+// List returns all webhooks for a scaling policy.
+func List(client *gophercloud.ServiceClient, groupID, policyID string) pagination.Pager {
+	url := listURL(client, groupID, policyID)
+
+	createPageFn := func(r pagination.PageResult) pagination.Page {
+		return WebhookPage{pagination.SinglePageBase(r)}
+	}
+
+	return pagination.NewPager(client, url, createPageFn)
+}
--- a/rackspace/autoscale/v1/webhooks/requests.go
+++ b/rackspace/autoscale/v1/webhooks/requests.go
@@ -153,3 +153,15 @@ func Update(client *gophercloud.ServiceClient, groupID, policyID, webhookID stri
 
 	return result
 }
+
+// Delete requests the given webhook be permanently deleted.
+func Delete(client *gophercloud.ServiceClient, groupID, policyID, webhookID string) DeleteResult {
+	var result DeleteResult
+
+	url := deleteURL(client, groupID, policyID, webhookID)
+	_, result.Err = client.Delete(url, &gophercloud.RequestOpts{
+		OkCodes: []int{204},
+	})
+
+	return result
+}
--- a/rackspace/autoscale/v1/webhooks/requests.go
+++ b/rackspace/autoscale/v1/webhooks/requests.go
@@ -78,15 +78,9 @@ func Create(client *gophercloud.ServiceClient, groupID, policyID string, opts Cr
 		return res
 	}
 
-	resp, err := client.Post(createURL(client, groupID, policyID), reqBody, &res.Body, nil)
+	_, res.Err = client.Post(createURL(client, groupID, policyID), reqBody, &res.Body, nil)
 
-	if err != nil {
-		res.Err = err
-		return res
-	}
-
-	pr := pagination.PageResultFromParsed(resp, res.Body)
-	return CreateResult{pagination.SinglePageBase(pr)}
+	return res
 }
 
 // Get requests the details of a single webhook with the given ID.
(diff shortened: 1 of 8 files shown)


jrperritt:
Changed the title to add the `[wip]` prefix. If/When this is ready for review, just change it to `[rfr]`.

rackspace/autoscale/v1/webhooks/results.go
```diff
@@ -0,0 +1,112 @@
...
+// Webhook represents a webhook associted with a scaling policy.
+type Webhook struct {
+	// UUID for the webhook.
+	ID string `mapstructure:"id" json:"id"`
+
+	// Name of the webhook.
+	Name string `mapstructure:"name" json:"name"`
+
+	// Links associated with the webhook, including the capability URL.
+	// TODO: Should this be []gophercloud.Link instead?
```
jrperritt: Yes, `[]gophercloud.Link`

rackspace/autoscale/v1/webhooks/results.go
```diff
@@ -0,0 +1,112 @@
...
+
+	return &response.Webhook, err
+}
+
+// CreateResult represents the result of a create operation. Multiple webhooks
+// can be created in a single call, so the result should be treated as a typical
+// pagination Page.  ExtractWebhooks() can be used to extract a slice of
+// Webhooks from a single page.
+type CreateResult struct {
+	pagination.SinglePageBase
```
jrperritt: By convention, only `List` results are pages. I know multiple webhooks can be returned here, but since it's only the ones that were just created, I think we're OK having a `gophercloud.Result` here.

rackspace/autoscale/v1/webhooks/results.go
```diff
@@ -0,0 +1,112 @@
...
+}
+
+// ExtractWebhooks extracts a slice of Webhooks from a CreateResult.
+func (res CreateResult) ExtractWebhooks() ([]Webhook, error) {
```
jrperritt: In line with the comment above, we can just have `Extract` here.

rackspace/autoscale/v1/webhooks/requests.go
```diff
@@ -0,0 +1,167 @@
...
+func (opts UpdateOpts) ToWebhookUpdateMap() (map[string]interface{}, error) {
+	if opts.Name == "" {
+		return nil, ErrNoName
+	}
+
+	hook := make(map[string]interface{})
+
+	hook["name"] = opts.Name
+
+	if opts.Metadata != nil {
```
jrperritt: This is why replacing entire configs during an update request is a bad idea. If a user wants to update just the webhook name, they'd have to 1) get the webhook metadata to supply to the `Update` call, and 2) know about 1). One thing we can do here is force users to supply a non-`nil` map for the metadata by returning an error if the `opts.Metadata` field is `nil`. In that case, to erase the current metadata, users would supply something like:
```go
err := webhooks.Update(autoscaleClient, webhooks.UpdateOpts{
    Name: "oldName",
    Metadata: make(map[string]string),
}).Extract()
```
Thoughts?

jrperritt:
This looks really good. Just a few comments.

rackspace/autoscale/v1/webhooks/results.go
bison: Cool. I think I was following the example of [`rackspace/lb/v1/nodes`](https://github.com/rackspace/gophercloud/tree/master/rackspace/lb/v1/nodes) here.

So, embed `gophercloud.Result` in place of `pagination.SinglePageBase` here, and rename the `ExtractWebhooks` method to just `Extract`, but still return a slice?

rackspace/autoscale/v1/webhooks/requests.go
bison: Agreed. That sounds good to me.

rackspace/autoscale/v1/webhooks/results.go
jrperritt: > Cool. I think I was following the example of rackspace/lb/v1/nodes here.

Hmm, that's a great point. And it so easily refutes everything I just claimed....
I'm still inclined to lean toward my original suggestion. To me, it's a bit unnatural to have users (know to) page through something other than a `ListResult`.

> So, embed gophercloud.Result in place of pagination.SinglePageBase here, and rename the ExtractWebhooks method to just Extract, but still return a slice?

Yes, exactly that.

bison: As far as I can tell, that's the only package that behaves that way, though I'm not sure how many of the APIs return lists in response to create or update operations.

Anyway, I agree this is kind of weird. I'll make it a `gophercloud.Result`.

rackspace/autoscale/v1/webhooks/results.go
bison: Done.

rackspace/autoscale/v1/webhooks/requests.go
bison: Done.

rackspace/autoscale/v1/webhooks/results.go
bison: Done.

jrperritt:
Looks great. +2

PR #550, opened 2016-04-01
[WIP] Add KeyName option for servers

hironobu-s:
This adds ```KeyName``` property to CreateOpts struct, and It will be able to create the instance with ssh public key.

https://developer.rackspace.com/docs/cloud-servers/v2/developer-guide/#post-create-server-servers


jrperritt:
In OpenStack, keypairs are an extension. Use the `keypairs` package: https://github.com/rackspace/gophercloud/blob/master/openstack/compute/v2/extensions/keypairs/requests.go#L14

hironobu-s:
Thank you for your advice!

PR #554, opened 2016-04-08
[RFR] Rackspace Auto Scale: policies

bison:
Adds a package for Rackspace Auto Scale [policies](https://developer.rackspace.com/docs/autoscale/v1/developer-guide/#document-api-operations/autoscale-policies).

### DIFF
--- a/rackspace/autoscale/v1/policies/requests.go
+++ b/rackspace/autoscale/v1/policies/requests.go
@@ -204,3 +204,15 @@ func Update(client *gophercloud.ServiceClient, groupID, policyID string, opts Up
 
 	return result
 }
+
+// Delete requests the given policy be permanently deleted.
+func Delete(client *gophercloud.ServiceClient, groupID, policyID string) DeleteResult {
+	var result DeleteResult
+
+	url := deleteURL(client, groupID, policyID)
+	_, result.Err = client.Delete(url, &gophercloud.RequestOpts{
+		OkCodes: []int{204},
+	})
+
+	return result
+}
--- a/rackspace/autoscale/v1/policies/requests.go
+++ b/rackspace/autoscale/v1/policies/requests.go
@@ -1,0 +1,17 @@
+package policies
+
+import (
+	"github.com/rackspace/gophercloud"
+	"github.com/rackspace/gophercloud/pagination"
+)
+
+// List returns all scaling policies for a group.
+func List(client *gophercloud.ServiceClient, groupID string) pagination.Pager {
+	url := listURL(client, groupID)
+
+	createPageFn := func(r pagination.PageResult) pagination.Page {
+		return PolicyPage{pagination.SinglePageBase(r)}
+	}
+
+	return pagination.NewPager(client, url, createPageFn)
+}
--- a/rackspace/autoscale/v1/policies/requests.go
+++ b/rackspace/autoscale/v1/policies/requests.go
@@ -9,8 +9,9 @@ import (
 
 // Validation errors returned by create or update operations.
 var (
-	ErrNoName = errors.New("Policy name cannot by empty.")
-	ErrNoArgs = errors.New("Args cannot be nil for schedule policies.")
+	ErrNoName            = errors.New("Policy name cannot by empty.")
+	ErrNoArgs            = errors.New("Args cannot be nil for schedule policies.")
+	ErrInvalidAdjustment = errors.New("Invalid adjustment type.")
 )
 
 // List returns all scaling policies for a group.
@@ -92,8 +93,9 @@ func (opts CreateOpts) ToPolicyCreateMap() ([]map[string]interface{}, error) {
 		policy["type"] = o.Type
 		policy["cooldown"] = o.Cooldown
 
-		// TODO: Function to validate and cast key + value?
-		policy[string(o.Adjustment.Type)] = o.Adjustment.Value
+		if err := setAdjustment(o.Adjustment, policy); err != nil {
+			return nil, err
+		}
 
 		if o.Args != nil {
 			policy["args"] = o.Args
@@ -176,8 +178,9 @@ func (opts UpdateOpts) ToPolicyUpdateMap() (map[string]interface{}, error) {
 	policy["type"] = opts.Type
 	policy["cooldown"] = opts.Cooldown
 
-	// TODO: Function to validate and cast key + value?
-	policy[string(opts.Adjustment.Type)] = opts.Adjustment.Value
+	if err := setAdjustment(opts.Adjustment, policy); err != nil {
+		return nil, err
+	}
 
 	if opts.Args != nil {
 		policy["args"] = opts.Args
@@ -227,4 +230,22 @@ func Execute(client *gophercloud.ServiceClient, groupID, policyID string) Execut
 	})
 
 	return result
+}
+
+// Validate and set an adjustment on the given request body.
+func setAdjustment(adjustment Adjustment, reqBody map[string]interface{}) error {
+	key := string(adjustment.Type)
+
+	switch adjustment.Type {
+	case ChangePercent:
+		reqBody[key] = adjustment.Value
+
+	case Change, DesiredCapacity:
+		reqBody[key] = int(adjustment.Value)
+
+	default:
+		return ErrInvalidAdjustment
+	}
+
+	return nil
 }
--- a/rackspace/autoscale/v1/policies/requests.go
+++ b/rackspace/autoscale/v1/policies/requests.go
@@ -9,11 +9,12 @@ import (
 
 // Validation errors returned by create or update operations.
 var (
-	ErrNoName            = errors.New("Policy name cannot by empty.")
-	ErrNoArgs            = errors.New("Args cannot be nil for schedule policies.")
+	ErrNoName            = errors.New("Policy name cannot be empty.")
+	ErrNoSchedule        = errors.New("Schedule cannot be nil for schedule policies.")
 	ErrCooldownRange     = errors.New("Cooldown is out of range (0, 86400).")
 	ErrUnknownType       = errors.New("Unknown policy type.")
 	ErrUnknownAdjustment = errors.New("Unknown adjustment type.")
+	ErrEmptyCron         = errors.New("Cron argument cannot be empty.")
 )
 
 // List returns all scaling policies for a group.
@@ -57,8 +58,9 @@ type CreateOpt struct {
 	// an integer.
 	AdjustmentValue float64
 
-	// Additional configuration options for some types of policy.
-	Args map[string]interface{}
+	// Value determining Schedule policy behavior, or nil for Webhook policies.
+	// This should be an appropriately configured Cron or an At value.
+	Schedule ScheduleArgs
 }
 
 // ToPolicyCreateMap converts a slice of CreateOpt structs into a map for use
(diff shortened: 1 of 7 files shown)


rackspace/autoscale/v1/policies/results.go
```diff
@@ -0,0 +1,138 @@
...
+	Name string `mapstructure:"name" json:"name"`
+
+	// Type of scaling policy.
+	Type Type `mapstructure:"type" json:"type"`
+
+	// Cooldown period, in seconds.
+	Cooldown int `mapstructure:"cooldown" json:"cooldown"`
+
+	// Number of servers added or, if negative, removed.
+	Change interface{} `mapstructure:"change" json:"change"`
```
bison: `Change`, `ChangePercent`, and `DesiredCapacity` are empty interfaces here because a policy can only have one of the three. The API will return whichever field is appropriate and omit the others. In reality, `ChangePercent` is a float and the other two are integers.

So, this lets you test if each is `nil`, though everything comes back as a float64. I could add an `UnmarshalJSON` method that converts to integers as needed. Is that necessary? Or is this off base altogether?

rackspace/autoscale/v1/policies/requests.go
```diff
@@ -0,0 +1,251 @@
...
+}
+
+// ToPolicyCreateMap converts a slice of CreateOpt structs into a map for use
+// in the request body of a Create operation.
+func (opts CreateOpts) ToPolicyCreateMap() ([]map[string]interface{}, error) {
+	var policies []map[string]interface{}
+
+	for _, o := range opts {
+		if o.Name == "" {
+			return nil, ErrNoName
```
bison: How much validation is appropriate here? I could also validate the range of values for `Cooldown` and that a known `Type` is provided, but maybe that should be left to the API.

rackspace/autoscale/v1/policies/results.go
jrperritt: I think you're on the right track here. I'd add an additional unmarshaling step: Define the `Policy` you have here as an unexported type (maybe in the `commonExtractPolicies` function as `policy`), decode the raw response data into that structure, and then add logic to massage it into a more user-friendly, exported `Policy` object like:
```go
type Policy struct {
//...
AdjustmentType AdjustmentType
AdjustmentValue float64
//...
}
``

File: rackspace/autoscale/v1/policies/requests.go [hunk 3]
```diff
@@ -0,0 +1,251 @@
...
+	Type Type
+
+	// Cooldown [required] period in seconds.
+	Cooldown int
+
+	// Adjustment [requried] type and value for the policy.
+	Adjustment Adjustment
+
+	// Additional configuration options for some types of policy.
+	Args map[string]interface{}
```
jrperritt: In the docs I only see `at` and `cron` as the possible values for `Args`. Are there more or is it just those? If it's just those I think we can define them concretely.

rackspace/autoscale/v1/policies/requests.go
jrperritt: As much as we can. We'd like to be able to:

- return early (without an HTTP call)
- provide our own error message (all the ones from the API will be 400s)

rackspace/autoscale/v1/policies/requests.go
```diff
@@ -0,0 +1,251 @@
...
+	Name string
+
+	// Type [required] of policy, i.e. either "webhook" or "schedule".
+	Type Type
+
+	// Cooldown [required] period in seconds.
+	Cooldown int
+
+	// Adjustment [requried] type and value for the policy.
+	Adjustment Adjustment
```
jrperritt: We can probably just embed the `AdjustmentType` and `AdjustmentValue` fields directly in here:
```go
type CreateOpt struct {
//...
AdjustmentType AdjustmentType
AdjustmentValue float64
//...
}
```

rackspace/autoscale/v1/policies/requests.go
bison: Yeah, just those AFAIK. Should this just be a `map[string]string` with some validation around what's provided? Or should it go further and define a `ScheduleArgs` type or something?

jrperritt: The latter. We'll want to be as specific as we can so that users know what their options are, and so that they don't have to worry about "cron" vs "CRON" vs "Cron" etc. Also, we can force the correct time format on users for the `At` field. Something like:
```go
type ScheduleType string
var (
  Cron ScheduleType = "cron"
  At ScheduleType = "at"
)

type CreateOpt struct {
//...
ScheduleType ScheduleType
ScheduleValue string
//...
}
```

That's ^ just an idea. There's an argument to made for defining it differently in such a way as to have the value for `At` be a `time.Time`, but having it as a string let's us check the format easily with `time.Parse`.

rackspace/autoscale/v1/policies/results.go
bison: Thinking about this more. Not sure this is doable with mapstructure. It's easy enough adding an `UnmarshalJSON` method to `Policy`, but I think that means doing one of the following:

- Keep the raw HTTP response in the `Body` field on the `gophercloud.Result` structs and do the decoding in the extract methods.

- Do the decoding in the request methods, and store a `Policy` in `gophercloud.Result.Body`. The extract methods don't do much then.

- Add a `Policy` to the `policyResult` struct and store the decoded policies there.

The first option seems most reasonable to me. Anyway, my hangup is that all of those break from the norm of decoding into a generic map on `gophercloud.Result.Body` and then using mapstructure. I can't find anywhere else in the code that does something like this.

Thoughts? Also, if it's easier, I can push something I have along the lines of the first option above and we can talk about it more concretely maybe.

jrperritt: > I guess you mean use mapstructure to decode into the unexported type then just define a helper to build a Policy from that. Is that right? 

Exactly. Doesn't have to be a separate function; could be defined in `commonExtractPolicies`.

rackspace/autoscale/v1/policies/requests.go
bison: Done.

rackspace/autoscale/v1/policies/results.go
bison: Makes sense. Done.

rackspace/autoscale/v1/policies/results.go
```diff
@@ -62,43 +66,97 @@ type ExecuteResult struct {
...
+	policy.Cooldown = p.Cooldown
+
+	policy.Args = p.Args
+
+	if v, ok := p.Change.(float64); ok {
+		policy.AdjustmentType = Change
+		policy.AdjustmentValue = v
+	}
+
+	if v, ok := p.ChangePercent.(float64); ok {
```
jrperritt: since only one of these can be set, we can `else if` here

rackspace/autoscale/v1/policies/results.go
```diff
@@ -62,43 +66,97 @@ type ExecuteResult struct {
...
+		policy.AdjustmentType = ChangePercent
+		policy.AdjustmentValue = v
+	}
+
+	if v, ok := p.DesiredCapacity.(float64); ok {
```
jrperritt: here, too

rackspace/autoscale/v1/policies/results.go
bison: Sure. Done.

rackspace/autoscale/v1/policies/requests.go
bison: Added validation for `Type` and `Cooldown`.

jrperritt:
Looks great. Let me know when you're done making changes and I'll merge it.

rackspace/autoscale/v1/policies/requests.go
bison: Playing around with a few options here, I had something more or less like this:

```go
type Policy struct {
  // ...
  Schedule ScheduleArgs
}

type ScheduleArgs interface {
  ToPolicyArgs() (string, string, error)
}

type At time.Time
type Cron string
```

Users could do a type switch on the `Schedule` field to determine which was returned. Thoughts?

jrperritt: Couldn't they switch on `ScheduleType`?

bison: Sure, you mean if the type and value are separate fields?

This just combines two steps. No need to set a string `ScheduleType` field if these types can handle that.

```go
Policy{
 Type: Schedule,
 Schedule: At(time.Now()),
}

Policy{
  Type: Schedule,
  Schedule: Cron("59 15 * * 0"),
}
```

Or am I misunderstanding?

jrperritt: Right, but instead of setting a string, we've got to define methods for `At` and `Cron` to implement the interface, and then users have to call `ToPolicyArgs` to get at the value. 

If you like, commit what you have in mind, and then a comment about how you'd expect users to use it.

bison:
Pushed 0310e5c with the `ScheduleArgs` interface. It's not finished. I haven't changed the create and update bits, but this works for getting / listing policies. Switching on the schedule type in those cases would look like this:

```go
policy, err := policies.Get(client, group, policyID).Extract()

switch s := policy.Schedule.(type) {
case policies.At:
        t := time.Time(s)
        fmt.Println("EXECUTE AT:", t.Format(time.RFC850))

case policies.Cron:
        fmt.Println("CRON:", s)

default:
        fmt.Println("Webhook.")
}
```

Creating policies with an `At` schedule seems nice with this because you can use all the standard time manipulation stuff and then create the `At` from that. Users shouldn't have to worry about the `ToPolicyArgs()` method. The methods for converting to maps for creates and updates can handle that.

bison:
Added the `ScheduleArgs` bits for create and update operations. Examples in the tests.

I think this is finished assuming you're good with the `ScheduleArgs` thing. If not, let me know and we can work out another way.

jrperritt:
Maybe we can compromise and just have `AtSchedule` and `CronSchedule` fields? After thinking about it, I don't think we should force users to switch (neither on type nor value) to figure out what's there.

bison:
We can do that. My only worry is the zero values. I really think `AtSchedule` works best as a `time.Time`, but if I request a policy configured with a cron schedule, I don't want `AtSchedule` set to `0001-01-01 00:00:00 +0000 UTC`... and in the opposite case, even having `Cron` be an empty string is a little strange.

We could make those fields `interface{}`, so they can default to `nil`, but that's not great either and it seems like users will have to do some checking to see what's what in any case.

bison:
Or did you mean keep the `ScheduleArgs` interface, but just have two fields and do some type validation before creates / updates?

jrperritt:
I'm very much a fan of the `ScheduleArgs in the `CreateOpts` and `UpdateOpts`; I think that works well. I trying to figure out if there's a way for users to get at that information in a response object without having to type-assert. Though I agree zero-values may not be the best solution either.

bison:
We could keep the `ScheduleArgs` interface and the two types for create and updates, and revert to `Args map[string]string` for `Policy`. If you request a policy with an `at` argument, then need to update it, you can always use `time.Parse` to turn it into a `policies.At`.

That doesn't seem ideal, and you still have to do some looking into the map, but maybe it's more straight forward than the type switch. I'm not really sure.

jrperritt:
No, let's go with what you have now. I think it's the best of the options I can think of right now.

jrperritt:
Is it ready to merge?

bison:
Yeah, I think it's good.

jrperritt:
+2

PR #556, opened 2016-04-14
[RFR] Rackspace Auto Scale: launch configurations

bison:
Adds a package for Rackspace Auto Scale [launch configurations](https://developer.rackspace.com/docs/autoscale/v1/developer-guide/#document-api-operations/configurations).

### DIFF
--- a/rackspace/autoscale/v1/launch/results.go
+++ b/rackspace/autoscale/v1/launch/results.go
@@ -2,6 +2,7 @@ package launch
 
 import (
 	"encoding/base64"
+	"encoding/json"
 	"reflect"
 
 	"github.com/mitchellh/mapstructure"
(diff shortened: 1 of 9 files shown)


rackspace/autoscale/v1/launch/results.go
```diff
@@ -0,0 +1,353 @@
...
+			file.Contents = []byte(c)
+		}
+	}
+
+	return &file, nil
+}
+
+// mapstructure.DecodeHookFuncType to convert from a string to []byte decoding
+// contents as base64 if possible.
+func stringToBytes(from reflect.Type, to reflect.Type, data interface{}) (interface{}, error) {
```
bison: This works well for decoding `user_data`, but it's unfortunate that mapstructure then seems to run the hooks for each resulting byte.

bison:
Finally got some time to finish this. Seems to be working for me.

PR #561, opened 2016-04-21
Add attach/detach blockstorage support

j-griffith:
Add volumeactions extension to blockstorage
module.  For now just start with basic attach
and detach commands and their associated
helpers.

Note that extensions are currently API version
independent in Cinder and are shared regardless
of configured Cinder API endpoint.

Add the following Cinder calls:
* reserve
* initialize_connection
* attach
* unreserve
* terminate_connection
* detach

For information on Cinder attach/detach workflow see:
http://docs.openstack.org/developer/cinder/devref/attach_detach_conventions.html

### DIFF
--- a/openstack/blockstorage/extensions/volumeactions/doc.go
+++ b/openstack/blockstorage/extensions/volumeactions/doc.go
@@ -1,0 +1,5 @@
+// Package volumeactions provides special actions on volumes as part of the
+// OpenStack Block Storage service.  Actions are the category of operations
+// that includes things like attach/detach of volumes to a host/instance.
+
+package volumeactions
--- a/openstack/blockstorage/extensions/volumeactions/requests.go
+++ b/openstack/blockstorage/extensions/volumeactions/requests.go
@@ -1,0 +1,154 @@
+package volumeactions
+
+import (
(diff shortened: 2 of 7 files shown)


j-griffith:
I missed the earlier work around V2 here:  https://github.com/rackspace/gophercloud/pull/489/commits/7516fde1624dab55990845c6f34f14b98f7dca56

I'll have a look at what I may have missed that's in there, or if that PR has a chance of being re-opened and reviewed I'm happy to help out.  I just need to get attach/detach somewhat soonish.

jamiehannaford:
@j-griffith I'm going to review 7516fde ASAP. At first glance, it seems that it provides what you need to get your Docker plugin to work. The only caveat is that it's been nested under a `v2` directory, which I'll ask the submitter to address. Do you know any docs that makes this version independent stuff explicit? I couldn't find any.

j-griffith:
Thanks Jamie, I'll load that up and give it a go.  Note I commented on that PR WRT the versions.  TLDR, no we don't really explicitly document that in Cinder, but the organization of the modules is fairly clear.  Also this will all change during the next six months on the cinder side and we will probably have to update/revisit here and impl micro versions.

PR #564, opened 2016-05-02
Extract additional info from GET

kaccardi:
Add some additional APIs to extract project, roles, and user information from a token retrieved with a GET request.  Move ExtractServiceCatalog to commonResult since it can be obtained either with a POST or a GET.

### DIFF
--- a/openstack/identity/v3/tokens/results.go
+++ b/openstack/identity/v3/tokens/results.go
@@ -102,7 +102,8 @@ func (r commonResult) ExtractToken() (*Token, error) {
 }
 
 // ExtractServiceCatalog returns the ServiceCatalog that was generated along with the user's Token.
-func (result CreateResult) ExtractServiceCatalog() (*ServiceCatalog, error) {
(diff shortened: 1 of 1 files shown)


jrperritt:
When this is ready for review, throw a `[rfr]` tag in front of the title and I'll have a look.

kaccardi:
ok, I modified the title, thanks.

openstack/identity/v3/tokens/results.go
```diff
@@ -46,6 +46,33 @@ type ServiceCatalog struct {
...
+}
+
+// RoleEntry contains the name of a user's role in a project.
+type RoleEntry struct {
+	Name string `mapstructure:"name"`
+}
+
+// Roles contains a list of role names extracted from a token.
+type Roles struct {
+	Entries []RoleEntry
```
jrperritt: Can you add a tag here?

jrperritt:
These look good. Can you add unit tests for them?

kaccardi:
yes, I will work on that.

PR #569, opened 2016-05-05
[WIP] Rackspace Auto Scale: groups

bison:
Adds a package for Rackspace Auto Scale [groups](https://developer.rackspace.com/docs/autoscale/v1/developer-guide/#document-api-operations/autoscale-groups).

### DIFF
--- a/rackspace/autoscale/v1/groups/fixtures.go
+++ b/rackspace/autoscale/v1/groups/fixtures.go
@@ -1,0 +1,236 @@
+// +build fixtures
+
+package groups
+
+import (
+	"fmt"
+	"net/http"
+	"testing"
+
+	"github.com/rackspace/gophercloud"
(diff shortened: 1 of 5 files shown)


bison:
Getting started on the final bit of adding Auto Scale support, but a few things are a little awkward already.

First, when you `GET` an individual group, the response consists of a UUID, configuration object, state object, launch configuration object, and a list of scaling policies. When listing groups, the groups only contain the UUID and the state object, so the others will be empty or `nil`. That seems kind of confusing, but I'm not sure what can be done about it.

Second, you don't update a group resource directly, you update its configuration at its `/config` URL, which you can also `GET` separately from the group. So, does it make more sense to have `GetConfig` and `UpdateConfig` (with an `ConfigUpdateOpts` type) methods in this package, or just have `groups.Update` apply to `/config`?

PR #573, opened 2016-05-13
[wip] Cinder v2

ncw:
This is a basic implementation of the cinder v2 API

The unit tests and the integration tests pass.

I've spread it out into lots of commits for ease of review - you might want them squashed.

I imagine there is stuff I've forgotten to change ;-)

My objective is to make terraform work with cinder/v2 api which it does on the bench with this patch.


ncw:
Looks like this has been merged already from #489 which I didn't see when creating this!

jamiehannaford:
@ncw Only the volumes resource was added along with the volume actions extension. We still need snapshots, volume types, etc. If you'd like to contribute these please feel free! I can review whenever you're ready

PR #574, opened 2016-05-14
[wip] Fixing v3 token auth

jtopjian:
This commit fixes identity v3 token authentication by looking at
the correct struct for the token. An acceptance test has been
added to verify token authentication works.

### DIFF
--- a/acceptance/openstack/identity/v3/token_test.go
+++ b/acceptance/openstack/identity/v3/token_test.go
@@ -5,6 +5,7 @@ package v3
 import (
 	"testing"
 
+	"github.com/rackspace/gophercloud"
 	"github.com/rackspace/gophercloud/openstack"
 	tokens3 "github.com/rackspace/gophercloud/openstack/identity/v3/tokens"
 )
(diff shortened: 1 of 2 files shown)


jtopjian:
I feel like the acceptance test is set up a little odd. It tries to make due with what devstack provides and accounts for missing info (like the Domain ID).

jtopjian:
Ah, I missed the unit tests in the `v3/tokens` package. This makes things very confusing...

The `AuthOptions` struct is able to take a [TokenID](https://github.com/rackspace/gophercloud/blob/master/auth_options.go#L54), and Identity v2 will look there to [authenticate via token](https://github.com/rackspace/gophercloud/blob/master/openstack/identity/v2/tokens/requests.go#L57-L60). But Identity v3 is looking at the [Client](https://github.com/rackspace/gophercloud/blob/master/openstack/identity/v3/tokens/requests.go#L87) for the token. I feel this is incorrect.

The POST requests in the API docs between [token authentication](http://developer.openstack.org/api-ref-identity-v3.html#authenticateTokenScoped) and [password authentication](http://developer.openstack.org/api-ref-identity-v3.html#authenticatePasswordScoped) look the same. Therefore, I feel the token should be referenced in `AuthOptions` and not the client. But maybe I'm incorrect...

@jrperritt thoughts?

jtopjian:
Closing in favor of #528 - totally missed that one...

PR #579, opened 2016-05-17
Fix selection of v2 volumes in openstack blockstorage

ncw:
After reviewing #489 I think this is a better way of selecting v2 volumes.  It certainly works in my testing.

This was committed originally in da546143

@feiskyer since you wrote the original code do you agree?

I'm unable to find the definitive openstack document which states valid values for the type field in the catalog though!


jrperritt:
At the least, the sample catalog [here](http://developer.openstack.org/api-ref-identity-v3.html#listServices) has the proposed fix as the type. +2

Issue #577, opened 2016-05-16
IdentityBase


suonto:
github.com/rackspace/gophercloud/openstack/client.go
Function: NewClient

Issue:
endpoint given is https://example.com/api/keystone/v3
IdentityBase resolves to https://example.com
It seems like IdentityBase should resolve to https://example.com/api/keystone, as endpoints are reconstructed like this: `v3Endpoint := client.IdentityBase + "v3/"`

Problem in action:
I'm running kubernetes with openstack provider and after on hour I'm starting to get errors like this:
```
openstack.go:284] Failed to find compute flavors: Error trying to re-authenticate: Expected HTTP response code [201 202] when accessing [POST https://example.com/v3/auth/tokens], but got 404 instead
```
I'm guessing the re-authentication is reconstructing identity endpoint from IdentityBase and fails because the path will be missing /api/keystone.

PR #582, opened 2016-05-23
Fix openstack NewClient to not remove relative paths from IdentityBase

suonto:
gophercloud.provider_client.go struct ProviderClient describes IdentityBase as
follows: "IdentityBase is the base URL used for a particular provider's
identity service - it will be used when issuing authenticatation requests. It
should point to the root resource of the identity service, not a specific
identity version."

Currently gophercloud.openstack.client.go func NewClient strips endpoints with
a non-empty relative path. For example, the IdentityBase returned for endpoint:

  http://example.com/foo

is:

  http://example.com/

when it should be:

  http://example.com/foo/

This change fixes NewClient to correctly handle such cases.

Unit tests are provided in openstack.test_client.go

[1] gophercloud.openstack.client.go

Fixes-Issue: #577

### DIFF
--- a/openstack/client.go
+++ b/openstack/client.go
@@ -3,6 +3,7 @@ package openstack
 import (
 	"fmt"
 	"net/url"
+	"regexp"
 	"strings"
 
 	"github.com/rackspace/gophercloud"
@@ -16,6 +17,8 @@ const (
 	v30 = "v3.0"
 )
 
+var VERSION_PATTERN = regexp.MustCompile("/v[1-9]{1}[0-9.]*")
+
 // NewClient prepares an unauthenticated ProviderClient instance.
 // Most users will probably prefer using the AuthenticatedClient function instead.
 // This is useful if you wish to explicitly control the version of the identity service that's used for authentication explicitly,
@@ -25,14 +28,18 @@ func NewClient(endpoint string) (*gophercloud.ProviderClient, error) {
 	if err != nil {
 		return nil, err
 	}
-	hadPath := u.Path != ""
-	u.Path, u.RawQuery, u.Fragment = "", "", ""
-	base := u.String()
 
+	u.RawQuery, u.Fragment = "", ""
+
+	// Base is url with path
 	endpoint = gophercloud.NormalizeURL(endpoint)
-	base = gophercloud.NormalizeURL(base)
+	base := gophercloud.NormalizeURL(u.String())
 
-	if hadPath {
+	var location = VERSION_PATTERN.FindStringIndex(base)
+
+	// If version found remove it from base
+	if location != nil {
+		base = base[0:location[0]+1]
 		return &gophercloud.ProviderClient{
 			IdentityBase:     base,
 			IdentityEndpoint: endpoint,
--- a/openstack/client.go
+++ b/openstack/client.go
@@ -3,7 +3,7 @@ package openstack
 import (
 	"fmt"
 	"net/url"
-	"regexp"
+	"strconv"
 	"strings"
 
 	"github.com/rackspace/gophercloud"
@@ -16,8 +16,6 @@ const (
 	v20 = "v2.0"
 	v30 = "v3.0"
 )
-
-var VERSION_PATTERN = regexp.MustCompile("/v[1-9]{1}[0-9\\.]{0,2}/")
 
 // NewClient prepares an unauthenticated ProviderClient instance.
 // Most users will probably prefer using the AuthenticatedClient function instead.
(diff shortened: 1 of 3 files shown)


jrperritt:
I think the current code is working as intended. If you know your identity endpoint, you can explicitly create a `ProviderClient` and set the `IdentityBase`/`IdentityEndpoint` fields

suonto:
Could you please share your reasoning supporting the claim that NewClient should strip the relative paths from IdentityBase? 

The justification for not stripping it is obvious: IdentityEndpoint is always constructed from IdentityBase + version. Thus, IdentityBase for endpoint http://example.com/foo/v3 should resolve to http://example.com/foo/, not http://example.com/.

Also, there was a similar relative path issue in python-openstacksdk, which was recognized and recently fixed:
https://github.com/openstack/python-openstacksdk/commit/1430a8f2479cc0cd1caf72cfaecfcd84925f1227

jrperritt:
>> Could you please share your reasoning supporting the claim that NewClient should strip the relative paths from IdentityBase?

I didn't claim that; I said it's working as intended. What's there works for default OpenStack implementations. Custom deployments can currently create a `ProviderClient` explicitly.

suonto:
I see. I also understand that you can create client and then modify the IdentityBase, but that applies only when you are using gophercloud directly. If you are, for example, running kubernetes with openstack cloud provider, it doesn't really work that way. Anyway, the new code works exactly the same as old for default endpoints (verified by unit tests), but also supports non-default endpoints. So, could you please merge it?

jrperritt:
I'll take a look on Monday

openstack/client.go
```diff
@@ -16,6 +17,8 @@ const (
 	v30 = "v3.0"
 )
 
+var VERSION_PATTERN = regexp.MustCompile("/v[1-9]{1}[0-9.]*")
```
jrperritt: Why is a regular expression needed here? We only have `v2.0` and `v3`.

suonto: You are right, the same functionality could be achieved with a simple if else structure. Another matter is, should the function return an error if a version other than "v2.0" or "v3" is found, or should it be stripped, or included as part of IdentityBase? My opinion is that we should check with regexp "/v[1-9]{1}[0-9.]{0,2}" and return an error if a match other than v2.0 or v3 is found. What you think?

jrperritt: I think remove the regex, because it's rarely what's needed. [At least] 2 options here:
1. extract the version, use a switch: cases 'v2.0', 'v3', 'default' . `default` returns error
2.  extract the version, strip the 'v', parse to float, check >= 2.0 and < 4.0, if not error

The first is the way to go if, as I understand it, Identity versions are static. OTOH, if users can have, say, Identity version 2.6 in the URL, then the second is the way to go.

suonto:
I preserved the regexp for a few reasons.
1. It's made for that
2. Alternative code that works as well looks quite ugly
3. I think that it's fundamentally better to find what we want and then process it, than to blindly take some part (for example the last piece of path) and then start checking if it is what we wanted
4. This is also extendable if there is ever a need for suffixes (path after matched version)

Added also more tests to verify thoroughly

jrperritt:
Regular expressions are lazy and error-prone. They are to be used when you don't know what you'll be parsing, nor where it will be. In this case, we know both the options and the location. Because of that, using a regular expression here is not appropriate. I'd gladly review and merge this PR if you want to implement [suggestion 1 in my comment above](https://github.com/rackspace/gophercloud/pull/582#discussion_r66099260]). Otherwise, it looks like we're at an impasse.

suonto:
@jrperritt have you had time to see the regexp-less implementation?

openstack/client_test.go
```diff
@@ -159,3 +159,95 @@ func TestAuthenticatedClientV2(t *testing.T) {
...
+	expected := fmt.Errorf("Invalid identity endpoint version %v. Supported versions: v2.0, v3", "v2.3")
+	th.AssertEquals(t, expected.Error(), err.Error())
+
+	_, err = NewClient("https://example.com/v2.0/foo")
+	th.AssertErr(t, err)
+	expected = fmt.Errorf("Path suffixes (after version) are not supported.")
+	th.AssertEquals(t, expected.Error(), err.Error())
+
+	// Does not match regexp, include to base
+	client, err = NewClient("https://example.com/v2a0/")
```
jrperritt: should fail

openstack/client_test.go
```diff
@@ -159,3 +159,95 @@ func TestAuthenticatedClientV2(t *testing.T) {
...
+	th.AssertNoErr(t, err)
+	th.AssertEquals(t, "", client.IdentityEndpoint)
+	th.AssertEquals(t, "https://example.com/v2a0/", client.IdentityBase)
+
+	// Does not match regexp, include to base
+	client, err = NewClient("https://example.com/v3api")
```
jrperritt: should fail

openstack/client.go
```diff
@@ -25,18 +26,40 @@ func NewClient(endpoint string) (*gophercloud.ProviderClient, error) {
...
+	path := u.Path
+	if !strings.HasSuffix(path, "/") {
+		path = path + "/"
+	}
+
+	parts := strings.Split(path[0:len(path)-1], "/")
+	for index,version := range(parts) {
+		if 2 <= len(version) && len(version) <= 4 && strings.HasPrefix(version, "v") {
+			_, err := strconv.ParseFloat(version[1:], 64)
+			if err == nil {
```
jrperritt: and if `err != nil` ?

suonto: Not a version, skipped like any other non-version part of the path like "foo".

openstack/client_test.go
suonto: In my opinion `/v2a0/`is a non-version path, and should be treated like `/foo/`. Now the question is, does the endpoint have to contain version, or should we treat the path as IdentityBase in case it doesn't. In the current proposal versionless endpoints are accepted and assigned as IdentityBase with an empty IdentityEndpoint.

jrperritt: I was thinking the version would be in the path, but I think you're right in thinking it may not be. I'm not sure though that we can have this test pass while the one at L242 fails.

suonto: This test, as well as the `/v3api/` test, are intended to serve as borderline cases where the path looks a lot like a version but is actually just a part of the path, like "foo". Currently, the distinction to version or non-version path is determined by a float conversion of the suffix following v, which is attempted only if the suffix length is in [1,3]. If you have a better idea how to do it, please share it.

jrperritt:
This looks good. Thank you for all the tests. +2

PR #586, opened 2016-05-24
Added cache-control header to objectstore

rodrigolourenco:
Added the header Cache-Control on objectstore.


rodrigolourenco:
@jrperritt I think this is ready for review, would you mind taking a look?

jrperritt:
+2

PR #587, opened 2016-05-26
[RFR] add admin actions

rtrox:
This PR adds all available admin actions to the "adminactions" extension.

API Docs for these functions are available here:
http://developer.openstack.org/api-ref-compute-v2.1.html#os-admin-actions-v2.1

The actions added are:
- createBackup
- injectNetworkInfo
- Migrate
- LiveMigrate
- ResetNetwork
- ResetState

### DIFF
--- a/openstack/compute/v2/extensions/adminactions/requests.go
+++ b/openstack/compute/v2/extensions/adminactions/requests.go
@@ -1,6 +1,10 @@
 package adminactions
 
-import "github.com/rackspace/gophercloud"
+import (
+	"fmt"
+
+	"github.com/rackspace/gophercloud"
+)
 
 func actionURL(client *gophercloud.ServiceClient, id string) string {
 	return client.ServiceURL("servers", id, "action")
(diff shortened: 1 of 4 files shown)


rtrox:
@jrperritt I think this is ready for review, would you mind taking a look?

rtrox:
I should add,  these features are being implemented in preparation for a PR to rack cli to allow admin functions by Rackspace admins. Thanks for looking!

jrperritt:
Great, I'll take a look tomorrow.

openstack/compute/v2/extensions/adminactions/requests.go
```diff
@@ -0,0 +1,134 @@
...
+	// Rotation: the number of backups to retain.
+	Rotation int
+}
+
+// ToBackupCreateMap assembles a request body based on the contents of a CreateOpts.
+func (opts CreateBackupOpts) ToCreateBackupMap() (map[string]interface{}, error) {
+	backup := make(map[string]interface{})
+
+	if opts.Name != "" {
+		backup["name"] = opts.Name
```
jrperritt: If `Name` is required, this should return an error if `opts.Name == ""`.

openstack/compute/v2/extensions/adminactions/requests.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+	if opts.BackupType != "" {
```
jrperritt: same thing here

openstack/compute/v2/extensions/adminactions/requests.go
```diff
@@ -0,0 +1,134 @@
...
+		backup["backup_type"] = opts.BackupType
+	}
+	backup["rotation"] = opts.Rotation
```
jrperritt: Is 0 allowed here?

rtrox: Based on the unit tests upstream, it looks like 0 is an acceptable rotation:

https://github.com/openstack/nova/blob/7d4ce9e73f75ac062652a6f500924b1bcec22991/nova/tests/unit/api/openstack/compute/test_create_backup.py#L189-L212

I'll add validation to ensure non-negative integers, though.

rtrox:
Thanks for the feedback @jrperritt !

I've addressed the line comment concerns by adding the following validation:

```go
	if opts.Name == "" {
		return nil, fmt.Errorf("CreateBackupOpts.Name cannot be blank.")
	}
	if opts.BackupType == "" {
		return nil, fmt.Errorf("CreateBackupOpts.BackupType cannot be blank.")
	}
	if opts.Rotation < 0 {
		return nil, fmt.Errorf("CreateBackupOpts.Rotation must 0 or greater.")
	}
	backup["name"] = opts.Name
	backup["backup_type"] = opts.BackupType
	backup["rotation"] = opts.Rotation
```

As I mentioned in my line comment, the rotation parameter can be zero according to the openstack unit-tests, but I added validation to ensure a non negative value (this is also in line with upstream validation).

I believe that means I am once more ready for review, would you mind taking another look?

jrperritt:
LGTM. +2

PR #588, opened 2016-06-01
Fix typos in "List Objects" doc

### LINKED ISSUE #568: List Objects documentation has some errors
I noticed a few issues with the code example on http://gophercloud.io/docs/object-storage/#list-objects

Currently, it shows:

### List objects
```
import "github.com/rackspace/gophercloud/pagination"

// We have the option of filtering objects by their attributes
opts := &objects.ListOpts{Full: true}

// Retrieve a pager (i.e. a paginated collection)
pager := objects.List(client, "container_name", opts)

// Define an anonymous function to be executed on each page's iteration
err := pager.EachPage(func(page pagination.Page) (bool, error) {

	// Get a slice of objects.Object structs
	objectList, err := objects.ExtractInfo(page)
	for _, n := range objectNames {
		// ...
	}

	// Get a slice of strings, i.e. object names
	objectNames, err := containers.ExtractNames(page)
	for _, n := range objectNames {
		// ...
	}

	return true, nil
})
```

---

#### Problem 1
```
	...
	// Get a slice of objects.Object structs
	objectList, err := objects.ExtractInfo(page)
	for _, n := range objectNames {
		// ...
	}
	...
```

This for loop is trying to loop over a variable that doesn't exist (it is declared a couple of lines further down). To me, this just looks like a mistaken copy&paste.

##### Solution
Changing `for _, n := range objectNames {` to `for _, o := range objectList {` fixes this piece of the example.

---

#### Problem 2
```
	...
	// Get a slice of strings, i.e. object names
	objectNames, err := containers.ExtractNames(page)
	for _, n := range objectNames {
		// ...
	}
	...
```
Running this results in the following error:
> panic: interface conversion: pagination.Page is objects.ObjectPage, not containers.ContainerPage [recovered]
	panic: interface conversion: pagination.Page is objects.ObjectPage, not containers.ContainerPage

##### Solution
So instead of using `containers.ExtractNames(page)`, using `objects.ExtractNames(page)` fixes this piece of the example.

---

#### Summary
I'd recommend changing that example to...

```
import "github.com/rackspace/gophercloud/pagination"

// We have the option of filtering objects by their attributes
opts := &objects.ListOpts{Full: true}

// Retrieve a pager (i.e. a paginated collection)
pager := objects.List(client, "container_name", opts)

// Define an anonymous function to be executed on each page's iteration
err := pager.EachPage(func(page pagination.Page) (bool, error) {

	// Get a slice of objects.Object structs
	objectList, err := objects.ExtractInfo(page)
	for _, o := range objectList {
		// ...
	}

	// Get a slice of strings, i.e. object names
	objectNames, err := objects.ExtractNames(page)
	for _, n := range objectNames {
		// ...
	}

	return true, nil
})
```

puddingfactory:
[Fix #568]
- 207: using previously declared "ObjectList" rather than "ObjectNames" which is for 212 instead.
- 212: using "objects.ExtractNames" rather than "containers.ExtractNames"

### DIFF
--- a/docs/object-storage/index.md
+++ b/docs/object-storage/index.md
@@ -204,12 +204,12 @@ err := pager.EachPage(func(page pagination.Page) (bool, error) {
 
 	// Get a slice of objects.Object structs
 	objectList, err := objects.ExtractInfo(page)
-	for _, n := range objectNames {
+	for _, o := range objectList {
 		// ...
 	}
 
 	// Get a slice of strings, i.e. object names
-	objectNames, err := containers.ExtractNames(page)
+	objectNames, err := objects.ExtractNames(page)
 	for _, n := range objectNames {
 		// ...
 	}


jrperritt:
Thank you. +2

PR #575, opened 2016-05-15
[rfr] Adding LBaaS v2 Support

dagnello:
This PR includes LBaaS V2 Support for the following components (ref. #460):
- Loadbalancers
- Listeners
- Pools and Members
- Health Monitors

LbaaS v2 API documentation:
http://developer.openstack.org/api-ref-networking-v2-ext.html
Note:  There are some details missing/invalid in current docs, I have been using client calls with --debug flag to see what the API actually returns.

PR #591, opened 2016-06-03
fixing build constraints for lbaas_v2 fixtures

sputnik13:
Build constraints are missing a line after the build constraint
as required by https://golang.org/pkg/go/build/, and the fixtures
are being pulled in by standard builds, causing test code to be
included in the final binary.


sputnik13:
@jrperritt @jamiehannaford this is related to https://github.com/rackspace/gophercloud/pull/575

we really need this fix in order to merge https://github.com/kubernetes/kubernetes/pull/25987

Issue #592, opened 2016-06-06
FEATURE FREEZE (Users of this library should read this)


jrperritt:
New features (read: features not currently pending in a PR) will no longer be added to this library. Over the next month (tentatively), the library will be migrated to [gophercloud/gophercloud](https://github.com/gophercloud/gophercloud) and the `rackspace` directory will be moved to its own, yet-to-be-made-public repository. Bug fixes will be reviewed and merged, and this repo will continue to exist, though it should be considered deprecated.

Pending pull requests in this repository will go through the review process and, if suitable, will ultimately be merged before being migrated to the new one(s). Outstanding issues that do not have pull requests will be addressed in the new repo(s). The new repos were created with the git history of this repo and as such user contributions will, for the most part, be maintained.

Currently the `gophercloud/gophercloud` repo is not accepting pull requests for 2 reasons:
1) the API is not completely stable, and
2) the contributing guide needs to be updated. Contributing to the new repo(s) will be a bit more structured so as to try to make PRs easier to review and therefore easier to merge.
When those 2 items are completed, this issue will be updated, and the repos will be open for PRs. Currently, issues may be created in `gophercloud/gophercloud` to discuss changes to the new API and feature requests.

The new repo(s) are almost entirely backwards-compatible with this one. Backwards-incompatible changes will be aggregated and enumerated here (in this issue's description). 

This issue will exist to provide updates on the migrations (ie. when the APIs are stable, when users can start contributing PRs to the new libraries, ...)

mikesimons:
Is this still on track? I see nothing in gophercloud/gophercloud since this issue was created but am holding off on progressing #590. I may just start against this repo and hope the swap is done by the time I'm complete :)

jrperritt:
Still on track

jrperritt:
Will be ready on Monday. I have changes locally that will be pushed upstream to `gophercloud/gophercloud`.

jrperritt:
Still on track to be done sometime today (Monday). Finishing porting LBaaS v2.

jrperritt:
The bulk of the work for this has been completed, at least to the point where issues can start being created against the [new repo](https://github.com/gophercloud/gophercloud). Please be sure to read the updated README, CONTRIBUTING and STYLEGUIDE documents before contributing.

[...]

PR #593, opened 2016-06-06
Improve identity v3 support

sbourlon:
OpenStack identity v3 instroduced domains which I added in this PR

- Add project domain and domain scoping
- Add user domain
- Add default domain in case the domain scope and the user domain are the same

Please, let me know if I am missing something or if this PR needs to be improved.

### DIFF
--- a/auth_options.go
+++ b/auth_options.go
@@ -27,10 +27,6 @@ type AuthOptions struct {
 	// preferred method of authentication.
 	Password, APIKey string
 
-	// At most one of DomainID and DomainName must be provided if using Username
-	// with Identity V3. Otherwise, either are optional.
-	DomainID, DomainName string
-
 	// The TenantID and TenantName fields are optional for the Identity V2 API.
 	// Some providers allow you to specify a TenantName instead of the TenantId.
 	// Some require both. Your provider's authentication policies will determine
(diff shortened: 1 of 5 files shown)


jrperritt:
Thank you for the PR, but this repo is currently in feature-freeze mode. See #592

sbourlon:
Thanks @jrperritt for the update. Would it still be possible to merge the PR after the feature-freeze mode is done?

jrperritt:
>> Would it still be possible to merge the PR after the feature-freeze mode is done?

In this repo, no. In `gophercloud/gophercloud`, yes. The public APIs are nearly identical and compatible with one another. As in, almost all code using this library can switch to using the new library with a single `gofmt` command replacing the import lines, and it should just work. For code that does require changes, I'd expect them to be minor ones.

sbourlon:
Thanks. I will test the PR against the new repo and sent a new PR to gophercloud/gophercloud.

jrperritt:
Please read the description of #592 in its entirety. The new repo won't accept PRs until the contributing guide is updated and the API stabilized. You can track issue #592 for updates on the migration and status of the new repo.

jrperritt:
Closing this. Hopefully we'll see it in the new repo

fatmcgav:
@sbourlon / @jrperritt Any joy with making these changes against gophercloud/gophercloud?

jtopjian:
@sbourlon @fatmcgav While reviewing https://github.com/hashicorp/terraform/pull/7041, I spent the past few evenings looking over the current gophercloud/gophercloud v3 implementation and I think everything is already in place. I even went as far as re-implementing the scoping portion of this PR and saw that the end result was the same.

While there is no direct support for `OS_PROJECT_DOMAIN_*`, `OS_USER_DOMAIN_*`, and `OS_DEFAULT_DOMAIN`, by using the correct combinations of `User*`, `Domain*`, and `Project*`, you can achieve the correct scoping.

_However_, I could be wrong. If so, can someone please provide an example of where scoping still needs work in the new Gophercloud repository?

fatmcgav:
@jtopjian I'm inclined to agree, as I've been running TF using the latest code and using domain and project scoping quite happily. 

Cheers 
Gav

PR #595, opened 2016-06-14
Support cinder volume create multiattach option

bergwolf:
both v1 and v2 cinder volume create api support multiattach option. Provide them so that callers can use it when necessary.


jrperritt:
Features are no longer being added to this library. See #592 
In a few weeks, you can make a PR to the `gophercloud/gophercloud` repo.

PR #596, opened 2016-06-18
fix cinder detach parameter

bergwolf:
It is "os-froce_detach" per http://developer.openstack.org/api-ref-blockstorage-v2.html

Signed-off-by: Peng Tao <bergwolf@gmail.com>


jrperritt:
`detach` is correct, as well. They are 2 different actions.

bergwolf:
hmm, just checked cinder code, there is indeed "os-detach" action.

Too bad only "os-force_detach" is documented in the cinder api reference page. Sorry for the noise.

jrperritt:
That's quite alright; it's reasonable to expect documentation. After working with OpenStack for a few years now, though, I've learned to go straight to the source like you eventually did.

PR #598, opened 2016-06-23
[rfr] Only pass volume_size if it is greater than zero

jtopjian:
A recent change in Nova now enforces volume_size to be greater than
zero. Since volume_size is an optional parameter, it should now only
be sent when it has a positive non-zero value.


jtopjian:
The change in Nova that caused this was implemented in both master and `stable/mitaka`, so this will affect people once they update to a forthcoming Mitaka point release.

jrperritt:
LGTM. +2

jtopjian:
Thank you!

PR #600, opened 2016-06-24
add Tenant ID option to UpdateOpts

sstrato:
Add Tenant ID field to UpdateOpts struct. 
It make possible to change Tenants ID of the neutron ports


jrperritt:
I'm not sure this is supported. Can you provide a URL that points to the line(s) in the Neutron source code that allows this?

PR #601, opened 2016-07-02
[rfc] Enable Debug Logging of Requests and Results

jtopjian:
This commit enables debug logs of the requests and results to be
printed if the GOPHERCLOUD_DEBUG environment variable is set.

### DIFF
--- a/provider_client.go
+++ b/provider_client.go
@@ -6,7 +6,9 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"log"
 	"net/http"
+	"os"
 	"strings"
 )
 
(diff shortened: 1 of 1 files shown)


jtopjian:
I was working with someone on troubleshooting a Terraform issue. When trying to compare the output of `nova --debug` or `openstack --debug` with Terraform / Gophercloud, we noticed there was no easy way to see what was being POST'd. The person I was helping actually used a network sniffer to pull the request 😄 

No doubt this isn't the best way to implement this, but thought I'd open the discussion and get some feedback.

jrperritt:
This should be handled by providing a custom `http.Client` to a `gophercloud.ProviderClient`. See an example of logging and limiting re-auth tries starting [here](https://github.com/rackspace/rack/blob/1.3/auth/clients.go#L309) and continuing to the end of the file.

jtopjian:
If I'm following correctly, any application that uses Gophercloud would need to implement a custom `http.Client` which provides the logging service?

jrperritt:
>> If I'm following correctly, any application that uses Gophercloud would need to implement a custom http.Client which provides the logging service?

Correct. We'd prefer not to dictate what should or shouldn't be logged, or which logger to use, or how many re-auth attempts are tried, etc.

jtopjian:
Got it. That makes sense. 

I would have never figured out how to implement logging outside of what I came up with. Maybe when the new repo is settled, some documentation or a reference logging client can be added?

I'll close this PR. Thank you for walking me through this :)

jrperritt:
>> Maybe when the new repo is settled, some documentation or a reference logging client can be added?

Yes, that's failure on my part.

PR #607, opened 2016-08-09
[rfr] Making SubnetID optional during pool and vip creation

uruddarraju:
At eBay, we have made changes to openstack to not have SubnetID mandatory during loadbalancer pool and vip creation. The LbaaS scheduler will find the right LB and the right subnet to create a vip and pool. And hence this can be optional.

This patch is needed to integrate Kubernetes with our internal private cloud LBaaS

cc @ashw7n @khgandhi


spothanis:
@jrperritt could you PTAL

jrperritt:
This repo is closed to changes. `gophercloud/gophercloud` is the new repo.

In addition, the options in this repo (and the new one) are for OpenStack proper, not customizations. Custom options may be passed to functions by satisfying the interface of the `*OptsBuilder`.

PR #609, opened 2016-08-10
Add openstack/auth_yaml.go

### LINKED ISSUE #487: Support clouds.yaml files
Keystone with auth plugins gets ... complex. As a result, openstack spawned the os-client-config library this last cycle, which, in addition to handling normal openrc env vars, allows the configuration of multiple clouds - and features of known providers in a yaml file:

http://docs.openstack.org/developer/os-client-config/

It would be super cool if gophercloud supported reading the same file. I'd be happy to work on adding the support - but I figured I should check first to see if it's a feature you'd take. I would imagine given the way env vars work - perhaps 'openstack.AuthOptionsFromYaml()' ?

raphink:
This aims to fix #487.

It is not totally finished, but I'd like some comments on it already.

/cc @emonty @mcanevet


jrperritt:
moved to here: https://github.com/gophercloud/gophercloud/pull/44

raphink:
Yes sorry for not closing this one.

PR #610, opened 2016-08-19
Fix generation of auth json map when using a trust id

MatMaul:
gofmt the code in a separate commit.

### DIFF
--- a/openstack/identity/v3/extensions/trust/request.go
+++ b/openstack/identity/v3/extensions/trust/request.go
@@ -7,35 +7,35 @@ import (
 )
 
 type AuthOptionsExt struct {
-  token3.AuthOptions
-  TrustID string
+	token3.AuthOptions
+	TrustID string
 }
 
 func (ao AuthOptionsExt) ToAuthOptionsV3Map(c *gophercloud.ServiceClient, scope *token3.Scope) (map[string]interface{}, error) {
 	//Passing scope value to nil to add scope later in this function.
 	authMap, err := ao.AuthOptions.ToAuthOptionsV3Map(c, nil)
-        if err != nil {
-                return nil, err
-        }
-  authMap = authMap["auth"].(map[string]interface{})
+	if err != nil {
+		return nil, err
+	}
+	authMap = authMap["auth"].(map[string]interface{})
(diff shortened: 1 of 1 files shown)


jrperritt:
@codevulture thoughts on this?

openstack/identity/v3/extensions/trust/request.go
```diff
@@ -7,34 +7,35 @@ import (
...
+		authMap["scope"] = map[string]interface{}{
+			"OS-TRUST:trust": map[string]interface{}{
+				"id": ao.TrustID,
+			},
+		}
+	} else {
+		return nil, token3.ErrScopeEmpty
 	}
-  return authMap, nil
+	return map[string]interface{}{"auth": authMap}, nil
```
codevulture: Funtion call --> ao.AuthOptions.ToAuthOptionsV3Map(c, nil)  would always a return an authMap of map[sring]interface type with "auth" as key as per: 
https://github.com/rackspace/gophercloud/blob/master/openstack/identity/v3/tokens/requests.go#L42
https://github.com/rackspace/gophercloud/blob/master/openstack/identity/v3/tokens/requests.go#L230
So is there a need for type assertion in L16 and this change?Please let me know if i am missing something? Thanks.

MatMaul: What do you mean ? If I remove the type assertion authMap is then of type interface{} after my first line of change, which means that I can't do authMap["scope"] = something a few lines later without triggering a compile error.

MatMaul: hum however I think we can safely remove the explicit cast on this line indeed. I'll try and push if it compiles.

MatMaul: Ok so it is not an explicit cast but an initialisation, I can't remove it, using {} doesn't initialise a map like in python.

codevulture:
@jrperritt, @MatMaul  : So, i tested this patch and this solves the problem of scope addition as mentioned.Thanks.

jrperritt:
@codevulture OK, thank you for reviewing and testing. +2

PR #614, opened 2016-09-20
Rename request_tests.go to request_test.go

MatMaul:
This is the usual convention and it is assumed by several tools like godep.

Needed for trust support in Kubernetes : 
https://github.com/kubernetes/kubernetes/pull/32111

### DIFF
--- a/openstack/identity/v3/extensions/trust/request_test.go
+++ b/openstack/identity/v3/extensions/trust/request_test.go
@@ -17,7 +17,7 @@ func authTokenPost(t *testing.T, options gophercloud.AuthOptions, scope *tokens.
 
 	client := gophercloud.ServiceClient{
 		ProviderClient: &gophercloud.ProviderClient{
-			TokenID: "12345abcdef",
+			TokenID: options.TokenID,
 		},
 		Endpoint: testhelper.Endpoint(),
 	}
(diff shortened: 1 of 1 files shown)


codevulture:
+1 godeps failure reason!

MatMaul:
Seems like the added tests were not call at all, because one is failing :) Next update should fix it.

strigazi:
@jrperritt Can you have a look at this?

MatMaul:
@jrperritt can we merge this please ? we are waiting for it to add trust id to Kubernetes. Thanks !

strigazi:
@jrperritt Do you have an update? We are blocked in kubernetes side.

openstack/identity/v3/extensions/trust/request_test.go
```diff
@@ -87,12 +66,3 @@ func TestTrustIDTokenID(t *testing.T) {
 
 	`)
 }
-
-func TestFailurePassword(t *testing.T) {
```
jrperritt: Don't delete tests; fix them instead.

MatMaul: This one never worked (those tests actually never ran before because of the filename), and doesn't seem really useful. I fixed the other one which was failing too. Be my guest if you understand and know how to fix this one.

jrperritt: Let me try again: If you want this PR merged, fix the unit test.

PR #617, opened 2016-10-19
detaching volume by attachment_id

jangseon-ryu:
This is a patch for detaching volume by attachment_id,
when using in attaching a same volume to multiple hosts.
This patch, Of course, support both 1:1 and 1:N attachment.

Currently, detaching logic in gophercloud can only support on 1:1(volume:host) attachment.
But many users want to be able to attach the same volume to multiple hosts.
(for example : clustered application(A:P))

BP : https://specs.openstack.org/openstack/cinder-specs/specs/kilo/multi-attach-volume.html

Signed-off-by: Jangseon Ryu <jangseon.ryu@navercorp.com>


j-griffith:
So technically this works, but it's pretty outdated.  The problem here is that the attachment_id isn't always accessible. I'd suggest at this point that we add the V3/Attachments controller and utilize that (as it's designed to do what you want here).  I have it on my list after Snapshots if that's helpful.

PR #642, opened 2018-05-08
Fixes problems with userdata on rackspace

gislifreyr:
When creating a server in the Rackspace public cloud the 'UserData'
option does not do anything unless the 'ConfigDrive' option is set to
'true'. This commit fixes that problem by checking if the UserData
option is non-empty and sets 'ConfigDrive' to 'true' if it is.


gislifreyr:
My first time with Go so if this is not up to standards then please excuse me and guide me to a better fix if possible.

